PR #18, opened 2020-09-18
Fixed typos and formatting

segfaultxavi:
Some simple typos and a bit of formatting to make it look nicer :)

I know one of the snippets is actually JSON, but I used YAML syntax highlighting because the snippet is not valid JSON (due to the ellipses). YAML is a superset of JSON so everything's fine.


fboucquez:
Thank @segfaultxavi . Everything looks perfect. You are a **writer** now. if you want to push to nemtech...

Issue #21, opened 2020-09-21
Allow "describe" command


fboucquez:
Add a new command that consoles out or export the configured properties for a given preset.

Example:

`symbol-bootstrap describe -p testnet -a dual`

Could render something like:

Identifier  | Value | Description
-- | -- | --
nodeEqualityStrategy  | host | Host vs PublicKey, this is for spam protection
nemesisSignerPublicKey  | 871B2F2F9D825252FFF43543066EE8D9141A3373F2F013BB449C6425A03362D8 | Which account will create nemesis transactions
generationHashSeed  | 1DFB2FAA9E7F054168B0C5FCB84F4DEB62CC2B4D317D861F3168D161F54EA78B | What will be the unique hash of the network for nodes joining

The output of this file could help maintain the docs keeping one source of truth for pages like:

https://docs.symbolplatform.com/guides/network/configuring-node-properties.html#properties

PR #26, opened 2020-09-29
Report command

fboucquez:
Hi @segfaultxavi @rg911 

I've added the "report" command that allows you to create RST files reporting the configured properties. Files are created in the `./target/report` folder

Examples:

```
symbol-bootstrap config --report 
symbol-bootstrap config --report -p testnet -a dual
symbol-bootstrap start --report
symbol-bootstrap report (after running a config)
```

An example of the generated report is the following:

https://github.com/nemtech/symbol-bootstrap/blob/config-report/test/expected-api-node-config.rst

The properties types and descriptions are kept in this [descriptions.yml](https://github.com/nemtech/symbol-bootstrap/blob/config-report/presets/descriptions.yml) file. The list is not comprehensive yet, we need to update the descriptions.yml file with all the possible keys. 

The idea is that you can generate the rst report files and copy them over symbol-docs to update guides like:

 https://docs.symbolplatform.com/guides/network/configuring-network-properties.html
 https://docs.symbolplatform.com/guides/network/configuring-node-properties.html

https://github.com/nemtech/symbol-docs/issues/516

- Added standalone command that generates rst files per node.
- Added --report to start and config commands
- Allowing API custom preset objects (instead of just a file).
- Added more customizable harvesting properties

Fixes https://github.com/nemtech/symbol-bootstrap/issues/21

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ $ npm install -g symbol-bootstrap
 $ symbol-bootstrap COMMAND
 running command...
 $ symbol-bootstrap (-v|--version|version)
-symbol-bootstrap/0.1.1 linux-x64 node-v12.18.4
+symbol-bootstrap/0.1.1 linux-x64 node-v14.8.0
 $ symbol-bootstrap --help [COMMAND]
 USAGE
   $ symbol-bootstrap COMMAND
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ $ npm install -g symbol-bootstrap
 $ symbol-bootstrap COMMAND
 running command...
 $ symbol-bootstrap (-v|--version|version)
-symbol-bootstrap/0.1.1 linux-x64 node-v14.8.0
+symbol-bootstrap/0.1.1 linux-x64 node-v12.18.4
 $ symbol-bootstrap --help [COMMAND]
 USAGE
   $ symbol-bootstrap COMMAND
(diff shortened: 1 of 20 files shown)


segfaultxavi:
This is awesome! Thanks a lot.
I want to discuss a few topics, though:

1. I would prefer that this command just output a simpler, machine-readable format, like CSV. In this way, it could be used in other places (to generate the OpenAPI descriptions, for example?) and I could adjust the format of the table in the docs without having to change symbol-bootstrap.
2. `descriptions.yml` looks like a very useful file. It seems a bit weird that it lives in this repo instead of some more important place. Anyway, we can move it later if other projects start using it.
3. `descriptions.yml` looks like a flat list. I thought the properties had a hierarchy? How do you group them when buidling the RST file?

fboucquez:
1. I've added CSV report next to RST. The RST is to simplify the symbol-docs update. We can also serialize the report as a Json file or a yml file if it's required.

This is an example of the CSV file https://github.com/nemtech/symbol-bootstrap/blob/config-report/test/expected-api-node-config.csv

2. If somebody defines something similar somewhere I'm happy to use it. I'm not super keen to maintain it. 

3. For simplicity yes. Descriptions and types are the same for properties using the same name in all the files. If for some reason, a property with a given name differs from the same name in a different file, we can improve file structure or prepend the file name to the property name in descriptions.yml.  The file and section structure is kept in RST and CSV files.

segfaultxavi:
I just discovered we already had a tool for this:
https://github.com/nemtech/catapult-docs-cli
We need to compare them and see which one is more convenient.

For now, it looks like `catapult-docs-cli` parses the catapult source files to extract all properties and descriptions, so we wouldn't need to keep a separate `descriptions.yml` (having two sources of truth). We would only need to extend `catapult-docs-cli` so that it also parses `symbol-bootstrap` to extract its own properties. How does this sound?

rg911 (commented):
looks good. 

just one thing I noticed, we should add open-source licences. probably can do it in another pr

fboucquez:
@segfaultxavi can you run catapult-docs-cli? how are the reports they provide? do you have examples? How's the rest report?
The command also bases the reports on the final .properties, not from the presets.
The tool is python, it may a bit harder to maintain than ts and it's probably just for us.  Bootstrap's command could be a more general command to (CSV exports for example)?

@rg911 TS license headers? I'll add them in another PR.  Thanks!

segfaultxavi:
The reports generated by `catapult-docs-cli` are the ones currently used in
https://docs.symbolplatform.com/guides/network/configuring-node-properties.html
and
https://docs.symbolplatform.com/guides/network/configuring-network-properties.html

I'm not saying we should use it, just that we should compare both tools, as you just did. All your points are valid, I agree with them.
However, I do like very much the idea that `catapult-docs-cli` does not depend on `descriptions.yml` which contains duplicated info. It directly parses the source files and extracts property names and descriptions straight from the origin.
Whatever tool we choose could do the same, and extract any bootstrap-specific properties by parsing `symbol-bootstrap` source, just like the other repos.

How does that sound?

fboucquez:
Mmmm. pulling the server code to extract descriptions it's a bit too convoluted for bootstrap. Can catapult-docs-cli generates a descriptions.yml like file and we copy it over?

segfaultxavi:
Sure, but then what would be bootstrap doing? just adding its own properties and then formatting the output? Isn't it simpler that docs-cli does all the work, pulling also from the bootstrap code? (it is already parsing catapult-server and catapult-rest)

fboucquez:
I've done a few more improvements removing the type and description columns when they are not provided. For now I'm going to maintain the description.yml file. We can improve later if needed.

Issue #29, opened 2020-10-01
Preset "repeat" to allow multiple service definitions


fboucquez:
Allow "repeat" field that generates multiple configuration for a service defined. The service could be a database, node or rest. For example:

`````
nemesis:
    transactionsDirectory: '/nemesis/transactions'
    binDirectory: /data/seed
    mosaics:
        - name: 'xym'
          divisibility: 6
          duration: 0
          supply: 8998999998000000
          isTransferable: true
          isSupplyMutable: false
          isRestrictable: false
          accounts: 5
databases:
    - repeat: 10
      name: 'db-$index'
      openPort: "27017 + $index" 
nodes:
    - repeat: 20
      harvesting: true
      voting: true
      api: false
      type: peer-node
      syncsource: true
      filespooling: false
      partialtransaction: false
      name: 'peer-node-$index'
      friendlyName: 'peer-node-$index'
      host: 'peer-node-$index'
      openPort: "7900 + $index" 
      sinkType: 'Sync'
      enableSingleThreadPool: true
      addressextraction: false
      mongo: false
      zeromq: false
    - repeat: 10
      harvesting: false
      voting: false
      api: true
      type: api-node
      syncsource: false
      filespooling: true
      partialtransaction: true
      friendlyName: 'api-node-$index'
      name: 'api-node-$index'
      host: 'api-node-$index'
      brokerHost: 'api-node-broker-$index'
      openBrokerPort: "8000 + $index"
      sinkType: 'Async'
      enableSingleThreadPool: false
      addressextraction: true
      mongo: true
      zeromq: true
      databaseHost: 'db-$index'
      enableAutoSyncCleanup: false
gateways:
    - repeat: 10
      apiNodeConfigPath: '/usr/local/share/symbol/api-node-config'
      apiNodeHost: 'api-node-$index'
      apiNodeBrokerHost: 'api-node-broker-$index'
      name: 'rest-gateway'
      description: 'catapult development network'
      maxConnectionAttempts: 7
      baseRetryDelay: 750
      databaseHost: 'db-$index'
      openPort: "3000 + $index"
networkType: 152
inflation:
    starting-at-height-1: 0
    starting-at-height-10000: 0
`````

Generates 20 peers 10 API, 10 databases, and 10 rest rest gateways. Note that `$index` will be automatically resolved with the index of the given instance. 

[...]

PR #36, opened 2020-10-05
Repeat and internal refactoring

fboucquez:
Hi team, I would like to show you a mid-size refactoring of the tool. All the changes are internal and they should not break the clients. This allows admins to scale up bootstrap. 

The goals:
* Allow scaling bootstrap to generate "100s" of nodes, gateways, databases, etc for new or existing networks.
* Improve the config to generate more reusable services. The `config` command is the main command. The `compose` and  `run` commands are there if the user wants to run the network using `docker-compose `. In the near future, we could add more commands to generate files for other services `cloud formation`, `kubernetes`, `terraform`, `ecs`, etc. I think the symbol specific generated configuration here can be merged with infrastructure configuration to generate a full cloud network. 

The changes:

**repeat**
This is the biggest feature. It allows you to "repeat" a given configuration N times. This is super useful to use when you want to generate the configuration for a new large network like testnet. Values in a repeat object support handlebars templating. You can use the $index to generate unique names, ports, or IP addresses. Index starts from 0.  @Wayonb @ravitej-aluru I got this feature from our last call. Please have a look.  

The default [bootstrap](https://github.com/nemtech/symbol-bootstrap/blob/preset-repeat/presets/bootstrap/network.yml#L35) setup now uses repeat. You can extend that one adding repeats to replicate multiple components. You can run a really big compose if you like!!! or just use the generated config. [Readme](https://github.com/nemtech/symbol-bootstrap/tree/preset-repeat#repeat-components) explains a little bit further.

Fixes https://github.com/nemtech/symbol-bootstrap/issues/29

**folder structure**
The biggest refactoring. I've changed the folder structure to be per service/component type instead of per bootstrap stage (config, data). See [readme](https://github.com/nemtech/symbol-bootstrap/tree/preset-repeat#target).
- The generated configuration of each component can be zipped and deployed into servers that may not use `compose` (like ECS or Kubernetes).
- The volumes in the compose are much simpler
- It allows relative folders in configuration
- `./target/addresses.yml` is the new file for the generated keys. It used to be `./target/config/generated-addresses/addresses.yml` @segfaultxavi 

**config**
- Generated configuration uses relative folders instead of absolute (example `dataDirectory: ./data`). This useful if you want to use the generated configuration for your barebone server. @Jaguar0625 @gimer
- /usr/catapult can be changed via preset if you have a different installation (also for running the server from barebone). 
- Dynamic default configuration. The default values of a node depend on the node type (the `harvesting:` `api:` `voting:` flags). See [here](https://github.com/nemtech/symbol-bootstrap/blob/preset-repeat/src/service/BootstrapUtils.ts#L172). This simplifies the out of the box presets. Users can still override those defaults, custom preset takes precedence like before. 
- nemesis block 1 is copied over to each node at config time, not when running docker-compose. This change is also to simplify the services' zip packaging and compose volumes

**compose**
- Allowed multiple database services (it was a bug)
- Custom bind ports can be used (example ` openPort: 3003`). If compose has multiple gateways like with repeat, you can bind them using ports 3001, 3002, 3003 etc (although internally all are 3000). 
- Allowing custom known peers. For example, in repeat, the api-X only allows rest-X
- Allowed custom IP addresses in services via preset.
- Allowed custom subnet via preset
- Simplified mounted volumes thanks to the new folder structure. 
- Simplified bash files. They don't need to copy block 1 and docker-compose uses working dir.
- Merged db and db-init services. Now the mongo database knows how to init itself. 

**resetData**
A new command that resets all the nodes data and database but keeping the generated configuration, keys, voting tree files, nemesis block, etc. When starting bootstrap, you can start from scratch, from the last configuration and new data, or just restart using the last configuration and the previous data. 

**misc**
- Removed dead code from other POCs
- Added a POC that generates docker images per generated node/service. This could be useful for cloud containers but not 100% yet.
- Added a lot of unit tests validating that the changes from this refactoring are the expected ones.


README.md
```diff
@@ -10,12 +10,12 @@ Symbol CLI tool that allows you creating, configuring and running Symbol&#39;s c
 [![Coverage Status](https://coveralls.io/repos/github/nemtech/symbol-bootstrap/badge.svg?branch=main)](https://coveralls.io/github/nemtech/symbol-bootstrap?branch=main)
 [![Api Doc](https://img.shields.io/badge/api-doc-blue.svg)](https://nemtech.github.io/symbol-bootstrap/)
 
-
 <!-- toc -->
 * [symbol-bootstrap](#symbol-bootstrap)
 * [Why this tool?](#why-this-tool)
 * [Key benefits:](#key-benefits)
 * [Concepts](#concepts)
+* [custom repeat preset for bootstrap preset.](#custom-repeat-preset-for-bootstrap-preset)
```
segfaultxavi, line 18:
```suggestion
```
If you link to this level-5 header you'll have to include links to all other level-5 headers :)

README.md
```diff
@@ -89,108 +88,142 @@ The best way to validate your configuration is by inspecting the generated confi
...
+
+`Repeat` is a pretty powerful customization. It tells bootstrap to "repeat" the defined configuration a number of times. This will allow you to generate as many services as you want quickly.
+
+This is specially useful for:
+
+-   Configure a really large network. The generated configuration could be then deployed into different cloud services.
+-   Load test many services together to validate how they behave.
+-   Exclude a service via `repeat: 0`.
+
+if repeat is active, each value in the object is [handlebars](https://handlebarsjs.com/guide/expressions.html) template. $index is index of the generated service starting from 0.
```
segfaultxavi, line 207:
```suggestion
If repeat is active, each value in the object is a [handlebars](https://handlebarsjs.com/guide/expressions.html) template. `$index` is the index of the generated service starting from 0. 
```

README.md
```diff
@@ -89,108 +88,142 @@ The best way to validate your configuration is by inspecting the generated confi
...
+
+This is specially useful for:
+
+-   Configure a really large network. The generated configuration could be then deployed into different cloud services.
+-   Load test many services together to validate how they behave.
+-   Exclude a service via `repeat: 0`.
+
+if repeat is active, each value in the object is [handlebars](https://handlebarsjs.com/guide/expressions.html) template. $index is index of the generated service starting from 0. 
+
+Given an $index of 10
```
segfaultxavi, line 209:
```suggestion
Given an `$index` of 10:
```

segfaultxavi (changes requested):
> `./target/addresses.yml` is the new folder for the generated keys. It used to be `./target/config/generated-addresses/addresses.yml` @segfaultxavi
Is this strictly necessary, specially two days after the docs have been released?

README.md
fboucquez: Let's rework the custom prests examples on https://github.com/nemtech/symbol-bootstrap/issues/37

rg911 (changes requested):

cmds/start-bootstrap-aws.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+./bin/run clean -t target/bootstrap
+./bin/run config -p bootstrap  -t target/bootstrap
+./bin/run compose -u current -r -t target/bootstrap --aws
```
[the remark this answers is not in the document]
fboucquez: Yeap, not working. I'll remove that file. I'll keep these unused lines https://github.com/nemtech/symbol-bootstrap/blob/f9c56bf7c57cdb4e8cfdbdde709b62082eee3e26/src/service/ComposeService.ts#L68 a little bit longer for reference.

cmds/convert-bootstrap.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+
+
```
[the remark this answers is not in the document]
fboucquez: Removed the file, convert file is from the aws POC.

fboucquez:
Hi team, this is a big PR. Could you check it out and run it from source code (npm install -g)?  Please play around and let me know what do you think. I'm pretty happy about it.

segfaultxavi:
You still haven't answered my question:
> `./target/addresses.yml` is the new folder for the generated keys. It used to be `./target/config/generated-addresses/addresses.yml` @segfaultxavi

Is this strictly necessary, specially two days after the docs have been released?

fboucquez:
> You still haven't answered my question:
> 
> > `./target/addresses.yml` is the new folder for the generated keys. It used to be `./target/config/generated-addresses/addresses.yml` @segfaultxavi
> 
> Is this strictly necessary, specially two days after the docs have been released?

I've missed that comment. Config folder doesn't exist anymore I prefer to have a simpler path for the addresses.yml. I'll create a PR in docs changing that folder we can merge around the same time I release this tool. This is not even a code-breaking change.

rg911:
> > You still haven't answered my question:
> > > `./target/addresses.yml` is the new folder for the generated keys. It used to be `./target/config/generated-addresses/addresses.yml` @segfaultxavi
> > 
> > 
> > Is this strictly necessary, specially two days after the docs have been released?
> 
> I've missed that comment. Config folder doesn't exist anymore I prefer to have a simpler path for the addresses.yml. I'll create a PR in docs changing that folder we can merge around the same time I release this tool. This is not even a code-breaking change.

so user with  config / target pre-generated won't be able to use the new version? seems to be a breaking change

segfaultxavi (dismissed):
My **documentation** concerns have been addressed. I'll update the docs site once this goes live.

fboucquez:
> 
> so user with config / target pre-generated won't be able to use the new version? seems to be a breaking change

Well, users would need to drop and recreate the network. They can provide the previously generated keys or presets stored in the current /config folder. I agree we should have a proper migration process but not even testnet has one. Should we explain the users (or script) how to move the folders around to reuse a previous state?

The tool started like a month ago from scratch. I reused some of the folder structure from the old bootstrap but they don't scale anymore when you want to create 100s of nodes or have folders per service ready to be cloud-deployed like we plan to do.

CHANGELOG.md
```diff
@@ -4,6 +4,27 @@ All notable changes to this project will be documented in this file.
...
+- **[BREAKING CHANGE]** Target folder structure has been change for scalability. The old target folder would need to be dropped when running this version. Backup the target folder if you need to keep your data!
+- **New Command:** `symbol-bootstrap resetData` it cleans the peer data and database without dropping the generated configuration.
+- Allowing `repeat` a node, a database or a gateway configuration multiple times. This allows you to create large network configuration quickly.
+- Path properties are now relative folder location. This improves reusability of the configuration when running the services outside docker compose.
+- Added node type based default configuration simplifying the configuration of nodes in presets.
+- Configurable `trustedHosts` and `localNetworks` in config.
+- Simplified mounted volumes in compose.
+- Allowed multiple databases in compose.
+- Compose's `openPort` now accepts port number.
+- Allowed custom ip address and subnet configuration in compose.
```
segfaultxavi, line 24:
```suggestion
- **[BREAKING CHANGE]** Target folder structure has been changed for scalability. The old target folder needs to be dropped when running this version. Backup the target folder if you need to keep your data!
- **New Command:** `symbol-bootstrap resetData` it cleans the peer data and database without dropping the generated configuration.
- Allowed `repeat` on a node, a database or a gateway to instantiate them multiple times. This enables you to create large network configurations quickly.
- Path properties are now relative folder locations. This improves reusability of the configuration when running the services outside docker compose.
- Added node type based default configuration simplifying the configuration of nodes in presets.
- Configurable `trustedHosts` and `localNetworks` in config.
- Simplified mounted volumes in compose.
- Allowed multiple databases in compose.
- Compose's `openPort` now accepts port number.
- Allowed custom ip address and subnet configuration in compose.
```

segfaultxavi (changes requested):

rg911 (approved):

Issue #54, opened 2020-10-26
Broker crashes while syncing from a new database


Wayonb:
**Describe the bug**
In the latest bootstrap, there is a race condition where mongors.sh script will exit without initializing the indexes for mongodb.  Catapult broker crashes due the indexes are not created correctly.  

The easiest fix it to remove the ` set -e` to allow the script to retry even when the mongod has not started yet.

https://github.com/nemtech/symbol-bootstrap/blob/47b2876432f795c13f8d7e2e12e0e099d54c5112/config/docker/mongo/mongors.sh#L2

Going forward, should investigate using the `docker-entrypoint-initdb.d` to initialize the mongodb.  I am not sure if its possible.
See the section on **Initializing a fresh instance**
https://hub.docker.com/_/mongo

PR #55, opened 2020-10-26
docker-entrypoint-initdb.d init scripts

fboucquez:
Fixes https://github.com/nemtech/symbol-bootstrap/issues/54


config/docker/mongo/mongoDbPrepare.js
```diff
@@ -1,80 +1,90 @@
+print(`DB Prepare`);
```
Wayonb, line 1: maybe I missed it, where do you specify the database to create the collections in?
There is an env MONGO_INITDB_DATABASE that should be set.

Wayonb (changes requested):

config/docker/mongo/mongoDbPrepare.js
fboucquez: You are right, I've just found out that the server inits the database itself, that's why it was working when I pushed the branch. Why is compose creating the tables if the server can do it? We would avoid maintaining the schemas on bootstrap.

fboucquez: it would also boot faster...

Wayonb: @Jaguar0625 @gimer Thoughts?

fboucquez:
Hi @Wayonb 

For now can use the copied db mongo script until we have server images with the scripts built-in. 

I've created the issue:

https://github.com/nemtech/symbol-bootstrap/issues/58

This issue depends on the server docker image to include the scripts.

Wayonb (approved):

PR #72, opened 2020-11-13
Initial 0.10.0.4 upgrade

Wayonb:
The initial 0.10.0.4 release with preset updated.  
Voting will not work until TS sdk is updated with the latest voting key changes.


config/node/resources/config-finalization.properties.mustache
```diff
@@ -11,4 +12,5 @@ messageSynchronizationMaxResponseSize = {{{messageSynchronizationMaxResponseSize
 
 maxHashesPerPoint = {{{toAmount maxHashesPerPoint}}}
 prevoteBlocksMultiple = {{{toAmount prevoteBlocksMultiple}}}
-votingKeyDilution = {{{votingKeyDilution}}}
+
+unfinalizedBlocksDuration = 0m
```
fboucquez, line 16: move value to shared.yml

config/node/resources/config-messaging.properties.mustache
```diff
@@ -1,3 +1,5 @@
 [messaging]
 
 subscriberPort = {{{subscriberPort}}}
+
+listenInterface = 0.0.0.0
```
fboucquez, line 5: move value to shared.yml

config/node/resources/config-node.properties.mustache
```diff
@@ -43,15 +43,19 @@ enableDispatcherInputAuditing = {{{enableDispatcherInputAuditing}}}
 maxCacheDatabaseWriteBatchSize = {{{maxCacheDatabaseWriteBatchSize}}}
 maxTrackedNodes = {{{toAmount maxTrackedNodes}}}
 
+minPartnerNodeVersion = {{{minPartnerNodeVersion}}}
+maxPartnerNodeVersion = {{{maxPartnerNodeVersion}}}
+
 # all hosts are trusted when list is empty
 trustedHosts = {{{trustedHosts}}}
 localNetworks = {{{localNetworks}}}
+listenInterface = 0.0.0.0
```
fboucquez, line 52: move value to shared.yml

config/node/resources/config-node.properties.mustache
```diff
@@ -43,15 +43,19 @@ enableDispatcherInputAuditing = {{{enableDispatcherInputAuditing}}}
...
 
 [localnode]
 
 host = {{{host}}}
 friendlyName = {{{friendlyName}}}
-version = 0
+version = 0.10.0.4
```
fboucquez, line 58: move value to shared.yml

fboucquez (changes requested):

fboucquez:
Thanks @Wayonb , ideally we should have all the values in the shared.yml by default. If a value is variable enough to be in a property file, it should be variable enough to be in shared.yml so node admin can change it with custom presets.

config/node/resources/config-messaging.properties.mustache
fboucquez: unsure if this is shared value with the one in config-node or not. If this value could be different, add a prefix to listenInterface like `messaginListenInterface` and `nodeListenInterface`

Wayonb:
@fboucquez Thanks for the feedback and fixes.  I fixed the tests :)

Issue #64, opened 2020-11-03
Error messages shortly after starting version 0.2.1 on newly installed Ubuntu20.04


garm01:
**error messages:**

- db exited with code 100

rest-gateway    | error: rest server is exiting due to error failed to connect to server [db:27017] on first connect [Error: getaddrinfo ENOTFOUND db
rest-gateway    |     at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {
rest-gateway    |   name: 'MongoNetworkError'
rest-gateway    | }] {"name":"MongoNetworkError","stack":"MongoNetworkError: failed to connect to server [db:27017] on first connect [Error: getaddrinfo ENOTFOUND db\n    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {\n  name: 'MongoNetworkError'\n}]\n    at Pool.<anonymous> (/app/catapult-rest/rest/node_modules/mongodb/lib/core/topologies/server.js:438:11)\n    at Pool.emit (events.js:315:20)\n    at /app/catapult-rest/rest/node_modules/mongodb/lib/core/connection/pool.js:562:14\n    at /app/catapult-rest/rest/node_modules/mongodb/lib/core/connection/pool.js:995:11\n    at /app/catapult-rest/rest/node_modules/mongodb/lib/core/connection/connect.js:32:7\n    at callback (/app/catapult-rest/rest/node_modules/mongodb/lib/core/connection/connect.js:280:5)\n    at Socket.<anonymous> (/app/catapult-rest/rest/node_modules/mongodb/lib/core/connection/connect.js:310:7)\n    at Object.onceWrapper (events.js:422:26)\n    at Socket.emit (events.js:315:20)\n    at emitErrorNT (internal/streams/destroy.js:92:8)","timestamp":"2020-11-03T13:36:28.799Z"}
api-node        | 2020-11-03 13:36:29.346079 0x00007fad0effd700: <info> (ionet::PacketSocket.cpp@41) eof reading from socket: End of file

- rest-gateway exited with code 0

- api-broker exited with code 253

- api-broker exited with code 253

**my steps after Install the OS:**(default configuration v.0.2.1 without a custom preset.yml)

apt install apt-transport-https ca-certificates curl software-properties-common gnupg-agent 
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
apt-key fingerprint 0EBFCD88
[...]

PR #76, opened 2020-11-22
Revisited lock files

fboucquez:
- Added --removeLocks
- Fixed runSeverRecover lock files location

@Wayonb , could we revisit the runServerRecover.sh file and how it's called? Currently, it doesn't do anything in dev/main as it's looking for the lock files in the wrong location. I've fixed this but now the server doesn't start correctly. Why is the recovery process executed in both server and broker docker containers? Should only one container do the recovery? Should one container (broker?) wait for the other container (server?) to finish booting/recovering?

Related to https://github.com/nemtech/symbol-bootstrap/issues/64

### DIFF
--- a/cmds/healthCheck-bootstrap.sh
+++ b/cmds/healthCheck-bootstrap.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
 set -e
 
-# docker rm -f $(docker ps -aq)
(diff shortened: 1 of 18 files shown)


fboucquez:
Superseded by https://github.com/nemtech/symbol-bootstrap/pull/166

PR #89, opened 2020-12-08
Docker compose params

fboucquez:
@Wayonb 

I've added a compose option to presets. You can add any compose option that will be injected into the compose service: For example a preset like:
````
nodes:
    - compose:
          cpu_count: 4
          shm_size: 64M
          deploy:
              resources:
                  limits:
                      memory: 4G
````

Will generate a service like:
````
    peer-node-0:
        user: '1000:1000'
        container_name: peer-node-0
        image: 'symbolplatform/symbol-server:gcc-0.10.0.4'
        command: bash -c "/bin/bash /symbol-commands/runServerRecover.sh  peer-node-0 && /bin/bash /symbol-commands/startServer.sh peer-node-0"
        stop_signal: SIGINT
        working_dir: /symbol-workdir
        restart: 'on-failure:2'
        ports:
            - '7900:7900'
        volumes:
            - '../nodes/peer-node-0:/symbol-workdir:rw'
            - './server:/symbol-commands:ro'
        depends_on: []
        cpu_count: 4
        shm_size: 64M
        deploy:
            resources:
                limits:
                    memory: 4G
````

Fixes https://github.com/nemtech/symbol-bootstrap/issues/19

### CODE REVIEW

File: presets/bootstrap/network.yml [hunk 1]
```diff
@@ -34,6 +34,11 @@ databases:
     - repeat: 1
       name: 'db-{{$index}}'
       openPort: '{{add $index 27017}}'
+      compose:
+        deploy:
+            resources:
```
fboucquez, line 39: @Wayonb could you confirm the best values for the out of the box docker attributes?

I'm wondering if we should keep it empty but add the extension `compose` point. So once we have a bit more flying time with the custom params we can make them official

Wayonb: yea, We can leave them empty for now.  At least they can be set if needed.

test/custom_compose_preset.yml
```diff
@@ -0,0 +1,16 @@
+databases:
+    - compose:
+          cpu_count: 2
+          mac_address: 02:42:ac:11:65:43
+          deploy:
+              resources:
+                  reservations:
+                      memory: 4G
+nodes:
```
Wayonb, line 9: Just curious.  What does reservation do?

Wayonb (approved):

test/custom_compose_preset.yml
fboucquez: I was expecting to start memory from this value (min) but it doesn't. Limit does limit the max value.

So, we are not 100% sure the best values so I would live it empty for now. I'll update

PR #94, opened 2020-12-17
Fix node link to node's public key

Wayonb:


fboucquez:
Hi Wayon,  thanks for the pr

SshpkService is not currently used. It's part of my attempt to remove using docker and openssl when generating the certificates. But I haven't got any luck yet.  The code is not used.

Also, the order of the fields in the interface doesn't affect the result.

```
        {
            privateKey,
            publicKey,
            key,
        };
```
 is a shortcut of:

```
        {
            privateKey: privateKey,
            publicKey: publicKey,
            key: key,
        };
```

Wayonb:
Thanks Fernando.  Yea, I realized my account is still incorrect. :)

Issue #108, opened 2021-01-06
Improve the recovery on the api node.


Wayonb:
**Describe the bug**
Currently, if the broker dies and recovery is run without the API node getting shutdown, it can lead to an issue with the DB.

**To Reproduce**
To reproduce the behavior:
kill the broker while the API node is running.

**Expected behavior**
if either the broker or API node crashes, both nodes need to stop then run the recovery tool before restarting.

fboucquez:
I don't know a way of doing this without merging broker and server in the same compose box.

fboucquez:
Fixed in dev

PR #123, opened 2021-01-19
Merged Broker/Server services into one

fboucquez:
Merged server and broker services into one service in docker-compose. 

Fixes https://github.com/nemtech/symbol-bootstrap/issues/108

Although broker and server are 2 individual processes, there are not really independent. If one crashes, the other one must be stoped and the recovery must run. It is really hard to have these 2 services working nicely in compose as both services need to communicate that the other's process has crashed and one must take the lead when restarting and running the recovery process. Recovery can only run in one of these services while the other is stopped. This makes the services synchronization harder. 

This PR simplifies the services, scripts, exception handling, and recovery by joining both services.

One related preset config is `dockerComposeServiceRestart: 'on-failure:2'`. This value fills the services' restart attribute. Possible values are https://docs.docker.com/config/containers/start-containers-automatically/

This value could be changed to `unless-stopped` via custom prest so if any service crashes unexpectedly, compose will restart it. This is especially important with the new broker/server service. When Docker restarts it automatically, the recovery will be executed before launching both processes.


fboucquez:
Superseded by https://github.com/nemtech/symbol-bootstrap/pull/166

PR #166, opened 2021-02-24
Broker/Server separate configs

fboucquez:
Splitting `userconfig` into `server-config` and `broker-config` for each service.
Improved server/broker start scripts.
Improved recovery? @Wayonb 

https://github.com/nemtech/symbol-bootstrap/issues/108


config/docker/server/startBroker.sh.mustache
```diff
@@ -1,19 +1,92 @@
 #!/bin/bash
```
Wayonb, line 1: Why combined all the scripts?

config/docker/server/startBroker.sh.mustache
```diff
@@ -1,19 +1,92 @@
...
+rm -f "{{{dataDirectory}}}/$application-recovery.started"
+
 if [ "$mode" == "DEBUG" ]; then
   echo "Setting up core dump..."
   mkdir -p ./logs
-  echo './logs/broker.%e.%t' > /proc/sys/kernel/core_pattern
+  echo "./logs/$application.%e.%t" >/proc/sys/kernel/core_pattern
+fi
+
+if [ "$waitForOther" == "true" ]; then
```
Wayonb, line 43: Is the wait needed for the broker?

config/docker/server/startServer.sh.mustache
```diff
@@ -1,17 +1,92 @@
...
+
 if [ "$mode" == "DEBUG" ]; then
   echo "Setting up core dump..."
   mkdir -p ./logs
-  echo './logs/server.%e.%t' >/proc/sys/kernel/core_pattern
+  echo "./logs/$application.%e.%t" >/proc/sys/kernel/core_pattern
 fi
 
-exec env LD_LIBRARY_PATH={{{catapultAppFolder}}}/lib:{{{catapultAppFolder}}}/deps {{{catapultAppFolder}}}/bin/catapult.server ./userconfig
+if [ "$waitForOther" == "true" ]; then
```
Wayonb, line 43: The wait should be after the recovery is run. If the other application has started the recovery will fail.

Wayonb (changes requested):

config/docker/server/startBroker.sh.mustache
fboucquez: I've merged together the recovery script and the start script for each service. I think it gives better and conrtol visibility. The service's run will stay the. Broker and Server scripts are pretty much the same, I'm thinking about merging them together into one.

config/docker/server/startBroker.sh.mustache
fboucquez: It's not needed. I've kept it so we can potentially have a common start.sh file.

config/docker/server/startServer.sh.mustache
fboucquez: Thanks, I'll move. it.
Question. Do you have a strategy about how to stop one service when the other one fails? If both catapult.broker and catapult.,server are running, and one stops. How to kill the other and clean up the flags so both services start (and recover) from scratch.

I'm thinking about pooling for the "theOtherApplication.started". It this file is removed (the other processed stoped), kill the current process. Do you have something better? Another alternative is using compose's health checks. But I don't want to ping the server.

Wayonb: Push some changes that will do the automatic recovery the api node.  Take a look....It works but needs some cleanup.

fboucquez: Thanks @Wayonb. I looks really nice.

I've merged the 2 scripts together into start.sh. It's not a mustache template anymore. I've added -f to the `rm` commands. The start.sh could be used potentially for non-docker-compose node admins. 

Could you validate https://github.com/nemtech/symbol-bootstrap/blob/broker_server_config/config/docker/server/start.sh

config/docker/server/start.sh
```diff
@@ -0,0 +1,127 @@
...
+touch "$dataDirectory/$application.started"
+
+processName="catapult.$application"
+echo "!!!! Starting $application...."
+$catapultAppFolder/bin/$processName "$config" &
+export applicationPid=$!
+
+cleanup() {
+  echo "Shutting down $application"
+  echo "kill -s SIGTERM $applicationPid"
```
Wayonb, line 97: missed removing the debug echo. -  `echo "kill -s SIGTERM $applicationPid"`

config/docker/server/start.sh
```diff
@@ -0,0 +1,127 @@
...
+    echo "The process $processName has already exited."
+    wait $applicationPid
+    exitStatus=$?
+    rm -f "./data/$application.started"
+    exit $exitStatus
+  fi
+
+  if [ -e $otherApplicationRecoveryFile ]; then
+    echo "Other process waiting for recovery. Shutting down $application"
+    echo "kill -s SIGTERM $applicationPid"
```
Wayonb, line 121: Here also.

Wayonb (dismissed):

PR #184, opened 2021-03-08
Custom Agent Port

fboucquez:
Agent port can be changed via `agentPort` custom preset property.


segfaultxavi:
Don't you want to talk about this new property in the README?

Issue #182, opened 2021-03-08
Local node should not be added to peers files.


Wayonb:
**Describe the bug**
When the peers file is generated the local node should not be added

Wayonb:
Example of what gets added for testnet nodes

    {
      "publicKey": "1650124BE4392427B52AE85AF1D45967D951BB3D413A3FF85BC5AC6263A1F015",
      "endpoint": {
        "host": "",
        "port": 7900
      },
      "metadata": {
        "roles": "Peer,Api"
      }
    },

fboucquez:
At all or only when there is no hostname/ip?

What about private networks? Each peer should be excluded from its own list? 
A network with peers A, B, C, 
A list should B,C, 
B list should have A, C, etc?

fboucquez:
Do we need peer files on the broker config folder?

PR #189, opened 2021-03-11
Removed node from its own peers json files.

fboucquez:
Fixes https://github.com/nemtech/symbol-bootstrap/issues/182


src/service/ConfigService.ts
```diff
@@ -349,15 +349,15 @@ export class ConfigService {
             addresses,
             serverConfig,
             NodeType.PEER_NODE,
-            (nodePresetData) => nodePresetData.harvesting,
+            (nodePresetData) => nodePresetData.harvesting && nodePresetData != nodePreset,
```
Wayonb, line 352: sorry, I should have explained this better in the bug.
Each node should have new preset indicating whether its a static or dynamic node.
mainnet nodes are dynamic except the core nodes which are static.
Private bootstrap you can leave those as static.

Wayonb (changes requested):

src/service/ConfigService.ts
fboucquez: If you are creating a new node, not an ngl node, should your node be on your list?

And in terms of bootstrap private, each node only has the other ones in this change. Is that ok? Or should each node have all the nodes, including themselves in the peer lists

PR #192, opened 2021-03-15
Ephemeral voting key

fboucquez:
Ephemeral Private Voting Key
Replaced PROMPT_MAIN_VOTING with PROMPT_MAIN_TRANSPORT
Improved logging and user feedback when asking keys and generating new voting key files


docs/presetGuides.md
```diff
@@ -280,6 +280,9 @@ Usage examples:
 
 If your custom preset contains private keys, it's highly recommended to `encrypt` it and provide `--password` when starting or configuring the node with Bootstrap.
 
+Note: Voting keys are ephemeral. They cannot be provided, bootstrap will always generate a new one when resetting the configuration. Bootstrap will only store the public key of the voting key for reference and linking.
+If the voting key file in the target folder gets lost, bootstrap will generate a new voting key file a new random key, and you would need to relink it to your main account.
```
segfaultxavi, line 284:
```suggestion
If the voting key file in the target folder is lost, bootstrap will generate a new voting key file from a new random key, and you will need to relink it to your main account (`link` command).
```

docs/presetGuides.md
```diff
@@ -328,7 +331,7 @@ If this is you case, you can tell Bootstrap to not store these keys with the `pr
...
 -   `ENCRYPT`: All private Keys are encrypted and stored in the target's `addresses.yml` file. Bootstrap will have them to be used when required. This is Bootstrap's default behaviour.
 -   `PROMPT_MAIN`: Main Private Keys are not stored in the target's `addresses.yml` file. Bootstrap will prompt for the Main Private Key when generating certificates, or transactions need to be signed in the `link` and `enrolProgram` commands.
--   `PROMPT_MAIN_VOTING`: Main and voting private keys are not stored in the target's `addresses.yml` file. Bootstrap will request the main private key when certificates are generated, or transactions need to be signed by the `link` and `enrolProgram` commands. The voting private key will be requested when generating the voting key file.
+-   `PROMPT_MAIN_TRANSPORT`: Main and transport private keys are not stored in the target's `addresses.yml` file. Bootstrap will request the main private key when certificates are generated, or transactions need to be signed by the `link` and `enrolProgram` commands. The transport private key would be asked when upgrading supernode agents.
```
segfaultxavi, line 334:
```suggestion
-   `PROMPT_MAIN_TRANSPORT`: Main and transport private keys are not stored in the target's `addresses.yml` file. Bootstrap will request the main private key when certificates are generated, or transactions need to be signed by the `link` and `enrolProgram` commands. The transport private key will be asked when upgrading supernode agents.
```

src/service/VotingService.ts
```diff
@@ -54,13 +53,11 @@ export class VotingService {
                 logger.info(`Voting File for node ${nodePreset.name} has been previously generated. Reusing...`);
                 return;
             }
+            const votingPrivateKey = nodeAccount?.voting.privateKey;
+            if (!votingPrivateKey) {
+                throw new Error('Voting key should have been previously generated!!!');
```
segfaultxavi, line 58: Can we tell the user what to do in this case?

segfaultxavi (changes requested):

src/service/VotingService.ts
fboucquez: Updated, atm they need to hard reset the target folder using the original custom preset. It should be an odd case where the voting key file got lost

Wayonb (approved):

PR #214, opened 2021-03-23
POC - Node certs in agent.

fboucquez:


fboucquez:
Not necessary after supernode certificate upgrade

PR #215, opened 2021-03-23
Peer role fix using syncsource

fboucquez:


src/service/ConfigLoader.ts
```diff
@@ -371,62 +380,60 @@ export class ConfigLoader {
...
+                enableSingleThreadPool: false,
+                addressextraction: true,
+                mongo: true,
+                zeromq: true,
+                enableAutoSyncCleanup: false,
+            };
+        }
+        //peer only (harvesting or not).
+        return {
+            sinkType: 'Sync',
```
Wayonb, line 419: The default sinkType should be Async

src/service/ConfigLoader.ts
```diff
@@ -371,62 +380,60 @@ export class ConfigLoader {
...
+            enableSingleThreadPool: true,
```
Wayonb, line 420: enableSingleThreadPool should be false

Wayonb (changes requested):

src/service/ConfigLoader.ts
fboucquez: Both changes are unrelated to the fix in this PR. 
syncType: Sync and enableSingleThreadPool: true has been there for a long time for peer-only nodes. Are we sure we want to apply this change now?

Wayonb: opened https://github.com/nemtech/symbol-bootstrap/issues/217

Wayonb (approved):

PR #230, opened 2021-04-06
Nodejs travis upgrade

fboucquez:
@rg911 we may need to do the same in the different projects. The first one on the list is important, is the one doing the special operations like release, alphas, docs, etc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
     - 12
     - lts/*
+    - node
     - 10
 services:
     - docker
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,8 +1,10 @@
 dist: bionic
 language: node_js
 node_js:
-    - 10
     - 12
+    - lts/*
+    - node
+    - 10
```
Wayonb, line 7: is nodejs v10 still needed?

Wayonb (approved):

.travis.yml
fboucquez: 10 is dropped

Issue #119, opened 2021-01-16
Verify Command: Version checking of required packages: node, docker and docker-compose


realgarp:
When setting up a new Symbol testnet node I ran into a version issue of docker-compose.
The default docker-compose on my Ubuntu 18.04 LTS was apparently not the version symbol-bootstrap expected, because docker-compose version complained about the format of the yml file generated by symbol-bootstrap. Trying to start the node with symbol-bootstrap does not work and exits in that case.
So I had to manually upgrade docker-compose to 1.27.4, to solve the issue.

The symbol-bootstrap docs mentions three required packages: node, docker and docker-compose.
It would be an improvement if symbol-bootstrap would be able to 
- check the installed versions of those three packages
- provide the docker and docker-compose versions also in the output of the -v command, as is already done for node.
- and if the required versions are not present: be able to indicate that to the user and not run at all before the user fixes it by upgrading the relevant package(s)

Thanks for considering... it would help future node owners.

segfaultxavi:
@fboucquez If you tell me the required versions for `docker` and `docker-compose` I'll add the info to the docs.

segfaultxavi:
I think I ran into the same issue on my Ubuntu 18.04, my `docker-compose` 1.25 crashed with:
```
AttributeError: module 'six' has no attribute 'ensure_text'
```
Is this the same thing you saw, @realgarp ?

segfaultxavi:
From a new folder:
```
$ docker --version
Docker version 19.03.8, build afacb8b7f0

$docker-compose --version
docker-compose version 1.25.0, build unknown

$ symbol-bootstrap --version
symbol-bootstrap/0.4.0 linux-x64 node-v15.0.1

$ symbol-bootstrap start -p bootstrap -a full -t test
                            _               _           _                       _           _                          
  ___   _   _   _ __ ___   | |__     ___   | |         | |__     ___     ___   | |_   ___  | |_   _ __    __ _   _ __  
 / __| | | | | | '_ ` _ \  | '_ \   / _ \  | |  _____  | '_ \   / _ \   / _ \  | __| / __| | __| | '__|  / _` | | '_ \ 
 \__ \ | |_| | | | | | | | | |_) | | (_) | | | |_____| | |_) | | (_) | | (_) | | |_  \__ \ | |_  | |    | (_| | | |_) |
[...]

PR #232, opened 2021-04-08
Added verify command

fboucquez:
Fixes https://github.com/nemtech/symbol-bootstrap/issues/119

Please have a look and let me know how it goes!

image


segfaultxavi:
It worked fine on my Windows machine (it detected no issues).
All texts look fine to me.

Maybe it would be worth mentioning that this only checks for SOFTWARE issues, not node configuration issues in any `target` folder. Otherwise, it might look like this performs a deep verification of your node configuration.

segfaultxavi (approved):

Issue #241, opened 2021-05-03
Use max fee of 0.1 Xym by default in enroll and link


fboucquez:
Atm the link and enroll commands use the node's min fee multipler. The result max fee is pretty low and the transactions are not confirmed. The current workaround is to use --maxFee 100000

This issue is to change the default max fee used to be 0.1 XYMs and remove the min fee multiplier.

PR #242, opened 2021-05-05
0.1 XYMs max fee default

fboucquez:
Replaced default Max Fee from the node's 'Min Fee Multiplier' to 0.1 XYMs.

Fixes https://github.com/nemtech/symbol-bootstrap/issues/241

### DIFF
--- a/presets/shared.yml
+++ b/presets/shared.yml
@@ -85,6 +85,7 @@ votingKeysDirectory: ./votingkeys
 enableDelegatedHarvestersAutoDetection: true
 catapultAppFolder: /usr/catapult
 rewardProgramAgentPort: 7881
+announceDefaultMaxFee: 100000
 
 # config database
 databaseName: catapult
--- a/src/service/AnnounceService.ts
+++ b/src/service/AnnounceService.ts
@@ -88,7 +88,7 @@ export class AnnounceService {
             description: 'If --ready is provided, the command will not ask for confirmation when announcing transactions.',
         }),
         maxFee: flags.integer({
-            description: 'the max fee used when announcing (absolute). The node min multiplier will be used if it is not provided.',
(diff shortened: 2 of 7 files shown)


presets/shared.yml
```diff
@@ -85,6 +85,7 @@ votingKeysDirectory: ./votingkeys
 enableDelegatedHarvestersAutoDetection: true
 catapultAppFolder: /usr/catapult
 rewardProgramAgentPort: 7881
+announceDefaultMaxFee: 100000
```
Wayonb, line 88: How did you reach that 0.1 as the default max fee?

src/service/AnnounceService.ts
```diff
@@ -265,32 +257,26 @@ export class AnnounceService {
...
                     );
-                    let lockFundsTransaction: Transaction = LockFundsTransaction.create(
+                    const lockFundsTransaction: Transaction = LockFundsTransaction.create(
                         deadline,
                         currency.createRelative(10),
                         UInt64.fromUint(1000),
                         signedAggregateTransaction,
                         networkType,
-                        defaultMaxFee,
+                        maxFee,
```
Wayonb, line 278: The max fee for a lock fund tx will be a lot smaller than aggregate tx.  Is there an issue with calculating the max fee from the minFee multiplier?

Wayonb (changes requested):

presets/shared.yml
fboucquez: It seems to be a sensible default. Cheap and fast enough.

src/service/AnnounceService.ts
fboucquez: The min fee multiplier was more problematic than expected. It creates a max fee too low that doesn't confirm. Maybe this is a bug in the server? Pretty much everyone is using the --maxFee param. 

I've seen this problem a lot https://forum.nem.io/t/symbol-supernode-programme-community-testing/30081/16

Issue #223, opened 2021-03-29
Add bootstrapVersion to rest.json


fboucquez:
Once https://github.com/nemtech/catapult-rest/issues/595 is finished

Issue #176, opened 2021-03-01
Unhandled exception when testing node/health during healthCheck


meyns:
Trying to run the healthCheck but apparently REST was not running fine yet and I got this exception:

2021-03-01T20:55:21.627Z info     Testing http://localhost:3000/node/health
(node:3362394) UnhandledPromiseRejectionWarning: #<Response>
(node:3362394) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:3362394) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
2021-03-01T20:55:21.643Z warn     Rest http://localhost:3000/node/health is NOT up and running YET: {"statusCode":503,"statusMessage":"Service Unavailable","body":"{\"status\":{\"apiNode\":\"down\",\"db\":\"up\"}}"}

fboucquez:
The health check should fail because Rest wasn't running. Is this happening?

The UnhandledPromiseRejectionWarning warning should be there.

meyns:
The rest was not running ok, but imo the unhandled exception should be handled much cleaner. Especially if it is deprecated.

PR #256, opened 2021-05-24
feat: bootstrap deployment metadata in rest /node/server

fboucquez:
fix: upgraded catapult-rest 2.3.6
fix: upgraded typescript sdk to 1.0.1
npm audit dependencies upgrade
Fixes https://github.com/nemtech/symbol-bootstrap/issues/223
Fixes https://github.com/nemtech/symbol-bootstrap/issues/176


test/reports/testnet-dual-voting/rest-gateway-rest.json
```diff
@@ -83,5 +83,10 @@
...
     }
   },
-  "numBlocksTransactionFeeStats": 300
+  "numBlocksTransactionFeeStats": 300,
+  "deployment": {
+    "deploymentTool": "symbol-bootstrap",
+    "deploymentToolVersion": "1.0.6",
+    "lastUpdatedDate": "2021-05-24"
+  }
 }
```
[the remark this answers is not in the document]
fboucquez: is lint complaining? these are test files to compare the generated with the expected

presets/shared.yml
```diff
@@ -211,6 +211,7 @@ maxConnectionAttempts: 15
 baseRetryDelay: 750
 connectionPoolSize: 10
 maxSubscriptions: 300
+restDeploymentTool: symbol-bootstrap
```
Wayonb, line 214: Why does ``restDeploymentTool`` needs to be update-able?  Can there be any other deployment tool?

Wayonb (changes requested):

presets/shared.yml
fboucquez: It's just the extension point just in case somebody for some reason wants to customize the string, same with the version and the date

Wayonb: Ok.  How can you tell if someone is using bootstrap when this is change?
If you want an extension point then I would think the bootstrap should always be present.  maybe like ``Bootstrap:<extension data>``

fboucquez: If the user really wants to mess up with the rest.json, he can manually update it after being generated. I'm just trying to not leave any place hardcoded

fboucquez: (There are other rest.json properties to be moved to shared, I would do at some point)

Wayonb (approved):

PR #263, opened 2021-06-07
WIP Ledger Integration

fboucquez:


docs/enrollRewardProgram.md
```diff
@@ -26,6 +26,9 @@ OPTIONS
 
   -u, --url=url                    [default: http://localhost:3000] the network url
 
+  --leger                          If --ledger is provided, the command will connect and sign the transactions with a
```
segfaultxavi, line 29:
```suggestion
  --ledger                         If --ledger is provided, the command will connect and sign the transactions using a
```

docs/link.md
```diff
@@ -24,6 +24,9 @@ OPTIONS
 
   -u, --url=url                    [default: http://localhost:3000] the network url
 
+  --leger                          If --ledger is provided, the command will connect and sign the transactions with a
```
segfaultxavi, line 27:
```suggestion
  --ledger                         If --ledger is provided, the command will connect and sign the transactions using a
```

segfaultxavi (changes requested):
Remember that you should always read everything before submitting :D

rg911 (commented):
@fboucquez  what's the rationale behind supporting ledger in bootstrap? Would it be better to create a separate tool for it?

docs/enrollRewardProgram.md
fboucquez: Thanks, this is an internal PR. I'm just sharing code with moha. Thanks for the fixes though...

fboucquez:
> @fboucquez what's the rationale behind supporting ledger in bootstrap? Would it be better to create a separate tool for it?

Bootstrap announces link and transfer transactions. This pr will let node admins sign those transactions with Ledger. 

When the main account is wealthy enough, it's highly recommended to convert it to a multisig (at least 1-1) where one of the cosigners is a Ledger.

I'm going to push a small typescript ledger client library that allows other CLIs (or wallets) to connect to ledger more easily. This would the first cli tool to accept ledger

PR #270, opened 2021-06-21
fix: custom preset cached on target's preset.yml

fboucquez:
This is a small improvement. The original custom preset (without any private key) is cached in the preset.yml of the target folder. If the user upgrading doesn't provide the --customPreset, that cache is used. Currently, if the user is not providing a custom preset when --upgrading, a custom value (for example maxUnlockedAccounts: 20) could be reset to its bootstrap's default (10).

--preset and --assembly are already derived from the target folder when not provided. 

How to upgrade a node before:
```
symbol-bootstrap start --upgrade --preset mainnet --assembly dual --customPreset my-custom-preset.yml
```
How to upgrade a node after:
```
symbol-bootstrap start --upgrade
```
The custom preset is loaded from the cache inside the target folder, the same applies to the preset and assembly values. 
The user can still provide a custom preset if he wants to change a property for example. 

Note: User needs to run the full --upgrade command one more time so the custom preset cache is populated.

Another fix: The --preset bootstrap is not the default anymore. Most people are creating nodes for testnet and mainnet, not for local testing networks like it was originally intended. The preset must be explicitly defined when starting a new node.


mm-s:
How does it behave when the custom preset contains private keys ?

fboucquez:
> How does it behave when the custom preset contains private keys?

Private keys are not stored in the ./target/preset.yml file, this files contains the catched custom preset. 

You can create your custom preset in your local offline machine, generate the configuration, zip the target folder, and send it to your node without the node having the keys at all.  With this PR, you don't need to send the custom preset either.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The changelog format is based on [Keep a Changelog](https://keepachangelog.com/e
 
 -   Added multi voting key file support.
 -   Added `updateVotingKeys` command.
+-   The `bootstrap` preset is not the default anymore. The name must be provided via --preset or as a custom preset field. 
+-   Added custom preset to the target folder. It's not required when upgrading the node without a configuration change.
```
segfaultxavi, line 13: It's **not** required? What do you mean?

src/service/ConfigLoader.ts
```diff
@@ -383,13 +383,10 @@ export class ConfigLoader {
...
-        const preset =
-            params.preset ||
-            params.customPresetObject?.preset ||
-            customPresetFileObject?.preset ||
-            oldPresetData?.preset ||
-            Preset.bootstrap;
-
+        const preset = params.preset || params.customPresetObject?.preset || customPresetFileObject?.preset || oldPresetData?.preset;
+        if (!preset) {
+            throw new KnownError('Preset value could not be resolved. Have you provided the --preset parameter?');
```
segfaultxavi, line 388:
```suggestion
            throw new KnownError('Preset value could not be resolved from target folder contents. Please provide the --preset parameter.');
```

segfaultxavi (changes requested):
I understand the purpose of this PR and I like it very much. But I do not understand that sentence in the README.

CHANGELOG.md
fboucquez: Soo, if you want to change a custom preset configuration, (like the friendly name), you need to provide the custom preset with --upgrade. If you are just upgrading the configuration, the custom preset is not required anymore. 

"The --customPreset param is only required when there is a configuration change, like a new 'friendlyName'"

src/service/ConfigLoader.ts
```diff
@@ -383,13 +383,10 @@ export class ConfigLoader {
...
         const oldPresetData = params.oldPresetData;
         const customPresetFileObject = this.loadCustomPreset(customPreset, params.password);
-        const preset =
-            params.preset ||
-            params.customPresetObject?.preset ||
-            customPresetFileObject?.preset ||
-            oldPresetData?.preset ||
-            Preset.bootstrap;
-
+        const preset = params.preset || params.customPresetObject?.preset || customPresetFileObject?.preset || oldPresetData?.preset;
```
Wayonb, line 386: Whats the order of precedence for the presets?  Should ``customPresetFileObject`` comes before ``customPresetObject``?

Wayonb (changes requested):

CHANGELOG.md
segfaultxavi: Ah, OK, understood. Then yeah, something like:

```suggestion
-   When upgrading, the --customPreset param is now only required if there is a configuration change, since
    the properties from the original custom preset are fully cached in the target folder.
```

But this is only the CHANGELOG so it's no big deal :D

segfaultxavi (dismissed):

src/service/ConfigLoader.ts
fboucquez: if the user is providing both, a file and a custom preset object, the custom preset object is priority. This is the case when you use bootstrap as a library. 

I would probably need to revisit this, especially in private networks when the network.yml is provided instead of built-in bootstrap.

fboucquez:
Superseded by https://github.com/nemtech/symbol-bootstrap/pull/277

fboucquez:
Hi @Wayonb I'm keen to merge this one. Is it ok for you?

docs/config.md
```diff
@@ -14,17 +14,19 @@ USAGE
   $ symbol-bootstrap config
 
 OPTIONS
-  -a, --assembly=assembly                   The assembly, example "dual" for testnet. If not provided, the value is
-                                            resolved from the target/preset.yml file.
+  -a, --assembly=assembly                   The assembly that define the node(s) layout. It can be provided via custom
```
Jaguar0625, line 17: that define*s

Jaguar0625: all three docs files. 

as there is a lot of similarity, i was wondering: are you generating these manually or automatically?

docs/start.md
```diff
@@ -79,11 +80,10 @@ OPTIONS
 
   --upgrade
       It regenerates the configuration reusing the previous keys. Use this flag when upgrading the version of bootstrap to 
-      keep your node up to date without dropping the local data. The original preset (-t), assembly (-a), and custom 
-      preset (-a) must be used. Backup the target folder before upgrading.
+      keep your node up to date without dropping the local data. Backup the target folder before upgrading.
 
 EXAMPLES
-  $ symbol-bootstrap start
+  $ symbol-bootstrap start -p bootstrap
```
Jaguar0625, line 86: delete? same as next line

README.md
```diff
@@ -66,9 +66,9 @@ Properties in each file override the previous values (by object deep merge).
 
 ### Out-of-the-box presets:
 
--   `-p bootstrap`: Default [preset](https://github.com/nemtech/symbol-bootstrap/blob/main/presets/bootstrap/network.yml). It's a private network with 1 mongo database, 2 peers, 1 api and 1 rest gateway. Nemesis block is generated.
+-   `-p bootstrap`: A [preset](https://github.com/nemtech/symbol-bootstrap/blob/main/presets/bootstrap/network.yml). It's a private network with 1 mongo database, 2 peers, 1 api and 1 rest gateway. Nemesis block is generated.
```
Jaguar0625, line 69: shouldn't this be a named preset ?

src/commands/pack.ts
```diff
@@ -112,7 +110,7 @@ export default class Pack extends Command {
         await ZipUtils.zip(targetZip, zipItems);
         await BootstrapUtils.deleteFile(noPrivateKeyTempFile);
         console.log();
-        console.log(`Zip file ${targetZip} has been created. You can unzip it in your node's machine and run:`);
-        console.log(`$ symbol-bootstrap start -p ${preset}${assembly ? ` -a ${assembly}` : ''} -c ${configOnlyCustomPresetFileName}`);
+        console.log(`You can unzip it in your node's machine and run:`);
```
Jaguar0625, line 113: i think it would be nice to output location of zip still

src/commands/start.ts
```diff
@@ -26,7 +25,7 @@ export default class Start extends Command {
     static description = 'Single command that aggregates config, compose and run in one line!';
 
     static examples = [
-        `$ symbol-bootstrap start`,
+        `$ symbol-bootstrap start -p bootstrap`,
         `$ symbol-bootstrap start -p bootstrap`,
```
Jaguar0625, line 29: remove duplicate line

src/commands/updateVotingKeys.ts
```diff
@@ -57,16 +55,24 @@ When a new voting file is created, Bootstrap will advise running the \`link\` co
...
-        const preset = existingPreset.preset;
-        if (!preset) {
-            throw new Error(`Network preset could not be resolved!`);
-        }
-        // Adds new shared/network properties to the existing preset. This is for upgrades..
-        const sharedPreset = BootstrapUtils.loadYaml(join(BootstrapUtils.ROOT_FOLDER, 'presets', 'shared.yml'), false);
-        const networkPreset = BootstrapUtils.loadYaml(join(BootstrapUtils.ROOT_FOLDER, 'presets', preset, 'network.yml'), false);
-        const presetData = configLoader.mergePresets(sharedPreset, networkPreset, existingPreset) as ConfigPreset;
 
+        const loadPresetData = () => {
```
Jaguar0625, line 59: why did you change this to a lambda instead of just being inline code?

src/service/ConfigLoader.ts
```diff
@@ -124,8 +127,12 @@ export class ConfigLoader {
 
         if (presetData.nemesis) {
             if (oldAddresses) {
+                if (!oldPresetData) {
+                    throw new Error('oldPresetData must be defined when upgrading!');
+                }
                 // Nemesis configuration cannot be changed on upgrade.
                 addresses.mosaics = oldAddresses.mosaics;
+                presetData.nemesis = oldPresetData.nemesis;
```
Jaguar0625, line 135: should you check that nemesis is the same ?

test/service/ConfigLoader.test.ts
```diff
@@ -61,6 +63,114 @@ describe('ConfigLoader', () => {
         expect(presetData).to.not.be.undefined;
     });
 
+    it('ConfigLoader custom maxUnlockedAccounts', async () => {
```
Jaguar0625, line 66: how is this test related to the rest of this change?

docs/config.md
fboucquez: Fixed. They are automatically generated

test/service/ConfigLoader.test.ts
fboucquez: It's testing that when recreating the `presetData` without providing the custom preset, the cached custom preset is used (and the user-provided maxUnlockedAccounts is kept for example).

src/commands/updateVotingKeys.ts
fboucquez: To avoid `*let* presetData: ConfigPreset;`, but it's an overkill. Lambda removed

README.md
fboucquez: I've updated the docs for this pr. 

I'm planning to change the assemblies, please have a look at the draft pr https://github.com/symbol/symbol-bootstrap/pull/328

Preview doc https://github.com/symbol/symbol-bootstrap/tree/shared_assemblies#preset

src/commands/pack.ts
fboucquez: Added

src/commands/start.ts
fboucquez: Removed

src/service/ConfigLoader.ts
fboucquez: I've removed that line. It's not currently required in this PR. 

I'm planning to provide a new PR improving how nemesis block is generated.
- Allow users to provide all the accounts, so the same custom preset executed twice (from scratch) generates the same accounts, balances, and nemesis block. 
- Allow users to include balances for both main and harvesting accounts (atm only the main mosaic can be 'opted-in')
- Validations that node's main accounts are not over the harvesting limit when dynamically generated (no user-provided)

src/service/ConfigLoader.ts
Jaguar0625: if you don't already, consider adding a build section to the readme (if you don't already) about how to generate these files and the expected test reporting files too. can be separate PR.

docs/config.md
Jaguar0625: if you don't already, consider adding a build section to the readme (if you don't already) about how to generate these files and the expected test reporting files too. can be separate PR.

Jaguar0625 (approved):

PR #281, opened 2021-07-21
HD/Mnemonic removed from wizard

fboucquez:
The current wizard in dev allows the user to enter the Mnemonic phrase to import keys. This was a good idea back when the wizard was created originally before mainnet launch. The original plan was to have people optin into secondary keys with seeds. This is not good anymore so I think it's safer to remove. Also, HD brings lots of dependencies that may cause issues.  If there is popular demand we can add it back but I haven't heard any request yet.


src/commands/wizard.ts
```diff
@@ -289,90 +284,42 @@ export default class Wizard extends Command {
...
-            }
-        }
-    }
-
-    private static async resolveAllAccount(
-        networkType: NetworkType,
-        derivedAccount: DerivedAccount | undefined,
-        rewardProgram: RewardProgram | undefined,
-    ): Promise<ProvidedAccounts> {
+    private static async resolveAllAccount(networkType: NetworkType, rewardProgram: RewardProgram | undefined): Promise<ProvidedAccounts> {
```
yilmazbahadir, line 287: Should we rename this method to `resolveAllAccounts`?

yilmazbahadir (approved):

PR #284, opened 2021-07-28
Added Native and Automatic HTTPS support to Rest, fixes #280

### LINKED ISSUE #280: Add HTTPS support to Rest
Related:

https://nemlog.nem.social/blog/58808
https://github.com/symbol/catapult-rest/issues/622
https://github.com/SteveLTN/https-portal

yilmazbahadir:
Fixes #280 

* Added HTTPS support selection to the wizard (Native(default) | Automatic | None)
    * Native HTTPS support selection will ask user to provide SSL key and certificate, file contents will be stored in the custom-preset file as base64 and encrypted(key), will be copied over the target/gateways folder during compose 
    * Automatic HTTPS support selection will result in setting `gateways[].httpsProxy(true|number)` which will integrate [Https-Portal](https://github.com/SteveLTN/https-portal) into the docker-compose.yml file during `compose`, default inbound port will be 3001
    * None: continue to have an insecure protocol as it used to be, HTTP 
* HttpsProxies preset is added to the ConfigPreset with the help of this preset user will be able to have more control over the parameters.
* Added unit tests


presets/testnet/network.yml
```diff
@@ -69,207 +69,182 @@ knownRestGateways:
...
-                roles: 'Peer,Voting'
-        -
-            publicKey: 2489946E49B03D9BE040E3FD42FEBC705D001A746BD25399E2796D615B35B732
-            endpoint:
-                host: ngl-peer-601.testnet.symboldev.network
-                port: 7900
-            metadata:
-                name: ngl-peer-601
-                roles: 'Peer,Voting'
+        - publicKey: 3CA5B729F46A56928C4971674B641C9535B93B560FC789F1134314786FC5766F
```
fboucquez, line 150: @yilmazbahadir have you excuted "npm run style:fix"? Maybe your IDE changed the yml format?

test/reports/bootstrap-voting/rest-gateway-0-rest.json
```diff
@@ -4,6 +4,9 @@
     "description": "catapult development network"
   },
   "port": 3000,
+  "protocol": "HTTP",
```
fboucquez, line 7: HTTP? Should it be HTTPS?

src/service/ConfigService.ts
```diff
@@ -692,6 +692,21 @@ export class ConfigService {
                     [],
                     ['node.crt.pem', 'node.key.pem', 'ca.cert.pem'],
                 );
+
+                if (gatewayPreset.restProtocol === 'HTTPS') {
+                    if (gatewayPreset.restSSLKeyBase64 && gatewayPreset.restSSLCertificateBase64) {
```
fboucquez, line 697: could you try what happens to the file when you --upgrade with and without providing the custom preset?

src/commands/wizard.ts
```diff
@@ -193,13 +196,40 @@ export default class Wizard extends Command {
...
+                symbolHostNameRequired ? 'This value is required when you are running on HTTPS!' : ''
             }`,
-            symbolHostRequired,
+            symbolHostNameRequired,
         );
+
+        let gateways;
+
+        if (httpsOption === HttpsOption.Native) {
+            // TODO ask keys and certs (file path or base64 encoded)
```
fboucquez, line 212: Remove TODO?

src/service/ComposeService.ts
```diff
@@ -293,6 +297,66 @@ export class ComposeService {
...
+                                if (httpsProxyPreset.domains) {
+                                    // domains already defined
+                                    if (httpsProxyPreset.domains.indexOf(n.name) < 0) {
+                                        // if rest-gateway not in the domains then add it
+                                        httpsProxyPreset.domains += ',' + restGatewayDomains;
+                                    }
+                                } else {
+                                    httpsProxyPreset.domains = restGatewayDomains;
+                                }
+                                if (typeof n.httpsProxy === 'number' && httpsProxyPreset.openPort === 3001) {
```
fboucquez, line 318: is  the "&& httpsProxyPreset.openPort === 3001" ok?

src/service/ComposeService.ts
```diff
@@ -293,6 +297,66 @@ export class ComposeService {
...
+            (presetData.httpsProxies || [])
+                .filter((d) => !d.excludeDockerService)
+                .map(async (n) => {
+                    const internalPort = 443;
+
+                    const volumes = [vol(`../${targetGatewaysFolder}/https-proxy`, nodeWorkingDirectory, false)];
+                    services.push(
+                        await resolveService(n, {
+                            container_name: n.name,
+                            // user, // TODO fix the user permissions issue
```
fboucquez, line 339: Does the user need Sudo to clean up the target folder? Are sudo files stored in the HTTP proxy volumen?

presets/shared.yml
```diff
@@ -71,6 +71,7 @@ symbolWalletImage: symbolplatform/symbol-desktop-wallet:1.0.1
 symbolFaucetImage: symbolplatform/symbol-faucet:1.0.1-alpha
 symbolAgentImage: symbolplatform/symbol-node-rewards-agent:2.0.0
 mongoImage: mongo:4.4.3-bionic
+httpsPortalImage: steveltn/https-portal:1
```
fboucquez, line 74: we would need an alpha rest eventually to test the native implementation

presets/testnet/network.yml
yilmazbahadir: Yeah, already fixed it but seems unpushed, pushing now.

test/reports/bootstrap-voting/rest-gateway-0-rest.json
yilmazbahadir: In the `shared.yml` file the protocol(restProtocol) is defaulted to HTTP and only if the HTTPS selection is `Native` on `wizard` command or manually changed to HTTPS in the custom-preset then it will be HTTPS.  But this should be the user's conscious decision I think.

src/service/ComposeService.ts
yilmazbahadir: Tried to check if the `openPort` in the `httpsProxies`is not altered and still stays as default then override it with the `httpsProxy` number value, makes sense?

presets/shared.yml
yilmazbahadir: Yes, I'm currently working on the `catapult-rest` part.

src/service/ComposeService.ts
yilmazbahadir: Not I'm afraid, I think it's because of the 's6-overlay' that the [https-portal docker](https://github.com/SteveLTN/https-portal/blob/master/Dockerfile) file is using. Couldn't find a decent solution yet but will continue to research; I think that docker file needs to do the setup with root user and then use a less privileged user. We might end up extending that docker file and keeping a less privileged user.

presets/shared.yml
yilmazbahadir: https://github.com/symbol/catapult-rest/pull/641

test/reports/bootstrap-voting/rest-gateway-0-rest.json
```diff
@@ -88,4 +91,4 @@
     "deploymentToolVersion": "abc",
     "lastUpdatedDate": "2021-05-23"
   }
-}
+}
```
Jaguar0625, line 94: [multiple files]
please make sure files end with a newline. you(r editor) seems to have truncated this in most of the json files.

src/service/ConfigService.ts
```diff
@@ -692,6 +692,30 @@ export class ConfigService {
...
+                            'base64',
+                        );
+                    } else {
+                        if (
+                            !existsSync(join(moveTo, presetData.restSSLKeyFileName)) &&
+                            !existsSync(join(moveTo, presetData.restSSLCertificateFileName))
+                        ) {
+                            logger.warn(
+                                `Native SSL is enabled but restSSLKeyBase64 or restSSLCertificateBase64 properties are not found in the custom-preset file! Either use 'symbol-bootstrap wizard' command to fill those properties in the custom-preset or make sure you copy your SSL key and cert files to target/gateways folder.`,
+                            );
```
Jaguar0625, line 711: should this abort?

src/service/ConfigService.ts
```diff
@@ -809,7 +833,10 @@ export class ConfigService {
         });
         (presetData.gateways || []).forEach(({ name }) => {
             const configFolder = BootstrapUtils.getTargetGatewayFolder(target, false, name);
-            BootstrapUtils.deleteFolder(configFolder);
+            BootstrapUtils.deleteFolder(configFolder, [
+                join(configFolder, presetData.restSSLKeyFileName),
+                join(configFolder, presetData.restSSLCertificateFileName),
```
Jaguar0625, line 838: why are you deleting these?

src/service/ComposeService.ts
gimer: `We might end up extending that docker file and keeping a less privileged user.` - probably, but given lack of `user` above, I'm guessing this should be solved sooner rather than later (read: prior to merging to `main`)

fboucquez:
@yilmazbahadir @Wayonb @gimer , I have drafted how the community can eventually upgrade the nodes to HTTPs using Bootstrap. 

https://hackmd.io/_F7gPWl2Td6WslNZRXeThA?view

segfaultxavi:
> @yilmazbahadir @Wayonb @gimer , I have drafted how the community can eventually upgrade the nodes to HTTPs using Bootstrap. Please have an initial look so then @segfaultxavi can do his magic
> 
> https://hackmd.io/_F7gPWl2Td6WslNZRXeThA?view

Yeah, ping me when you're done reviewing... and when you give me access to the file :)

fboucquez:
Moved to syndicate HackMD https://hackmd.io/Cx1KQCOxRaaHrNNwuNI7CQ?view

presets/mainnet/assembly-dual.yml
```diff
@@ -20,3 +20,11 @@ gateways:
...
       openPort: true
       ipv4_address: 172.20.0.25
+httpsProxies:
+    - name: 'https-proxy'
+      excludeDockerService: true #disabled as default
+      openPort: 3001
+      #domains: 'symbol-node.example.com -> http://rest-gateway:3000'
+      stage: 'production'
+      webSocket: 'true'
+      serverNamesHashBucketSize: 128
```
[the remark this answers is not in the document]
yilmazbahadir: This is an assembly-specific configuration, so we decided to have them in the relevant assembly yaml files only. If there is no `rest-gateway` in the assembly then this configuration is irrelevant, that's why it's copied over relevant assembly files(`dual` and `api`) and excluded by default (user needs to decide which option he wants to have, `native | automatic-letsencrypt | none`)

fboucquez (changes requested):

fboucquez:
I just realized that my comments were pending for the last 2 weeks. I just submitted the comments. Sorry guys.

src/commands/wizard.ts
```diff
@@ -133,9 +138,7 @@ export default class Wizard extends Command {
 
         if (!flags.skipPull) {
             const service = await new BootstrapService();
-            console.log();
-            console.log('Pulling catapult tools image before asking to go offline...');
-            console.log();
+            console.log('\nPulling catapult tools image before asking to go offline...\n');
```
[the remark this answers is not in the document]
yilmazbahadir: I think the idea was to make that particular message more visible. So it's just a refactor, instead of using `console.log()` before and after, used new lines.

presets/shared.yml
```diff
@@ -214,3 +215,9 @@ connectionPoolSize: 10
 maxSubscriptions: 300
 restExtensions: 'accountLink, aggregate, lockHash, lockSecret, mosaic, metadata, multisig, namespace, receipts, restrictions, transfer, cmc'
 restDeploymentTool: symbol-bootstrap
+restProtocol: HTTP
```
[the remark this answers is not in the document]
yilmazbahadir: `restProtocol` is the protocol of the REST compose service. We don't know which option the user will take to enable HTTPS at this point. If the let's encrypt option is selected, `restProtocol` would still be HTTP because the HTTPS protocol will be covered by the reverse-proxy(https-portal).

presets/shared.yml
```diff
@@ -214,3 +215,9 @@ connectionPoolSize: 10
...
+restSSLPath: '/symbol-workdir'
+restSSLKeyFileName: 'restSSL.key'
+restSSLCertificateFileName: 'restSSL.crt'
+restSSLKeyBase64: ''
```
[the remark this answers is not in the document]
yilmazbahadir: OK, removed.

src/service/ComposeService.ts
yilmazbahadir: I removed the user and the volume parts from `ComposeService`. The container will hold the certificates produced unless there is an image upgrade which doesn't happen frequently, so we are safe from the produced file permissions issues. There will be another PR for the `symbol-bootstrap stop` command improvement (which will prevent containers to be removed when stopped).

Jaguar0625 (approved):

PR #287, opened 2021-08-13
fix: bootstrap stop command is changed to run  <docker-compose stop> instead of <docker-compose down>

yilmazbahadir:
<docker-compose down> stops and removes the containers thus the files produced in the container get lost on each stop command.
We changed it to <docker-compose stop> in order to preserve the last state of the container and have a faster start-up.


fboucquez (changes requested):
Nice one! Could you add a comment to the changelog.md?

yilmazbahadir:
`CHANGELOG.md` is updated.

fboucquez (approved):

PR #289, opened 2021-09-11
fix: `updateVotingKeys` NAN fix when upgrading bootstrap cli

fboucquez:
How to replicate:

The user has bootstrap 1.0.6 installed and running.  The user installs bootstrap 1.0.7 and runs `updateVotingKeys` without `--
upgrade` the node. The current target folder created with 1.0.6 doesn't have `votingKeyDesiredLifetime` as it's a new property from 1.0.7. 

This PR hotfixes the issue.

Question, should we name this release 1.0.7-hotfix or 1.0.8? 

Note that we have alphas with 1.0.8 already, a 1.0.8 released hotfix will have fewer features than the alphas. New alphas would be 1.0.9-alpha. It's not a big issue as you need to specify the alpha version or alpha tag to get an alpha. Regular installs will use released versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The changelog format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [1.0.8] - Sep-13-2021
+
+**Milestone**: Mainnet(1.0.1.0)
+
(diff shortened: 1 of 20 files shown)


cryptoBeliever:
In my opinion new version with hotfix should be 1.0.8 and new alpha 1.0.9

Wayonb (approved):

PR #291, opened 2021-09-13
fix: isHostValid method refactor, subdomain and domain names can start with number, test cases added

### LINKED ISSUE #288: 1.0.8-alpha "wizard" If the first character is a number when entering a domain string, it will not be accepted.
In 1.0.8-alpha-202108171450
The domain/IP setting using "wizard"
ex: "dusan.gq" This will pass
However
ex: "3.dusan.gq" This is rejected

If enter a domain / IP with a number at the beginning, it will be rejected if it is regarded as a domain.
If it is considered that you entered the IP, it will pass

yilmazbahadir:
fixes #288


fboucquez (approved):
Awesome! thanks @yilmazbahadir

fboucquez (approved):
Awesome! thanks @yilmazbahadir

PR #299, opened 2021-10-10
feat: operating account (optional) parameter is added to link command

### LINKED ISSUE #290: Is there any method to execute `symbol-bootstrap link` command remotely?
**Is your feature request related to a problem? Please describe.**
Some `symbol-bootstrap` commands , such as `symbol-bootstrap link`,  are not assumed that a node operator is not the private key holder of the node. 

This makes it difficult to operate nodes by other persons and/or entities; node operators are  professional server engineers but do not want to know the private key or other holder's asset secret information, although `symbol-bootstrap link` command requests to input holder's private key.

**Describe the solution you'd like**
If symbol-bootstrap command has `remote sign option`, it would work well.

1. (node operator) symbol-bootstrap command export transaction before signing in text or qr code, and pass it to the holder.
2. (holder) sign the transaction via wallet.

**Describe alternatives you've considered**
Or node operator execute an aggregate tx equal to execute `symbol-bootstrap link` command, when the main account is a multisig account.

yilmazbahadir:
fixes #290 

This feature will allow node operators to announce aggregate transactions on behalf of the node owners(main) account without needing main account private key.

image


src/service/AnnounceService.ts
```diff
@@ -221,161 +229,170 @@ export class AnnounceService {
...
+                    cosigners,
+                    generationHash,
+                    currency,
+                    transactionService,
+                    listener,
+                    ready,
+                    nodeAccount.name,
+                );
+                if (!announced) {
+                    continue;
```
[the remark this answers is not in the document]
yilmazbahadir: fixed

src/service/AnnounceService.ts
```diff
@@ -141,31 +156,49 @@ export class AnnounceService {
             }
             const nodePreset = (presetData.nodes || [])[index];
             const mainAccount = PublicAccount.createFromPublicKey(nodeAccount.main.publicKey, presetData.networkType);
+            if (operatingPublicKey) {
+                logger.info(
+                    `The Operating Account[public key: ${operatingPublicKey}] is creating transactions on behalf of your node account[public key: ${mainAccount.publicKey}].  Signers and cosigners may see a warning when signing the transactions on the Wallets!`,
```
fboucquez, line 161: Could we unify the format in a method that knows how to print a (public)account? So we then define if we want to show public key, address or both in a centralized place. 

Even though address can be derived from a public key, addresses are easier to see and most people wouldn't be doing that public key -> address conversion

src/service/AnnounceService.ts
```diff
@@ -221,161 +229,155 @@ export class AnnounceService {
...
-                    );
-                    let lockFundsTransaction: Transaction = LockFundsTransaction.create(
+                    logger.info(`Cosigner ${bestCosigner.address.plain()} is initializing the transactions.`);
+                    signerAccount = bestCosigner; // override with a cosigner when multisig
+                    requiredCosignatures = multisigAccountInfo.minApproval;
+                }
+                const mainMultisigAccountInfo = await this.getMultisigAccount(repositoryFactory, mainAccount.address);
+                requiredCosignatures += mainMultisigAccountInfo?.minApproval || 0; // mainAccount.minApproval
+
+                const zeroAmountInnerTransaction = (account: Account): Transaction =>
```
fboucquez, line 268: should receive a PublicAccount

src/service/AnnounceService.ts
```diff
@@ -221,161 +229,155 @@ export class AnnounceService {
...
-                            ', ',
-                        )}. The tool will ask for the cosigners provide keys in order to announce the transactions. These private keys are not stored anywhere!`,
-                );
-                const cosigners = await this.promptAccounts(
+            const cosigners: Account[] = [];
+
+            if (operatingPublicAccount) {
+                // ask for operatingAccount private key
+                const operatingAccount = Account.createFromPrivateKey(
+                    await CommandUtils.resolvePrivateKey(
```
fboucquez, line 237: Really nice the way you also accept mutlsig operator accounts. It could happen that both operator and main accounts are multisig, so the bonded would be accepting/asking cosignatures from both.

Only ask for the operator private key when the operator account is not multisig, move the prompt to the else of `if (multisigAccountInfo) {`

src/service/AnnounceService.ts
```diff
@@ -487,4 +489,215 @@ export class AnnounceService {
...
+    }
+
+    private async getMultisigBestCosigner(
+        msigAccountInfo: MultisigAccountInfo,
+        cosigners: Account[],
+        accountName: string,
+        networkType: NetworkType,
+        repositoryFactory: RepositoryFactory,
+        currencyMosaicId: MosaicId | undefined,
+    ) {
```
fboucquez, line 687: Add return type

test/service/AnnounceService.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    AnnounceService,
+    BootstrapService,
+    CommandUtils,
+    ConfigService,
+    Preset,
+    RemoteNodeService,
+    RepositoryInfo,
+} from '../../src/service';
+
+describe('Announce Service', () => {
```
fboucquez, line 29: Thanks for the unit tests!

PR #307, opened 2021-10-30
feat: convert to or modify multisig account added

yilmazbahadir:
`symbol-bootstrap modifyMultisig`  command is added. It's useful to convert main account into a multisig account or to modify the multisig account structure. Takes the following parameters, the user will be prompted for the unset parameters.

```
-A, --addressAdditions=addressAdditions              Cosignatory accounts addresses to be added (separated by a
                                                       comma).

  -D, --addressDeletions=addressDeletions              Cosignatory accounts addresses to be removed (separated by a
                                                       comma).

  -a, --minApprovalDelta=minApprovalDelta              Number of signatures needed to approve a transaction. If the
                                                       account is already multisig, enter the number of cosignatories to
                                                       add or remove.

  -r, --minRemovalDelta=minRemovalDelta                Number of signatures needed to remove a cosignatory. If the
                                                       account is already multisig, enter the number of cosignatories to
                                                       add or remove.
```


src/service/AnnounceService.ts
```diff
@@ -358,18 +358,39 @@ export class AnnounceService {
...
-                            providedMaxFee,
-                            minFeeMultiplier,
-                            generationHash,
-                            currency,
-                            transactionService,
-                            listener,
-                            ready,
-                            nodeAccount.name,
-                        );
+                        if (transactions[0] instanceof MultisigAccountModificationTransaction) {
```
rg911, line 361: would it be better to use the transactionType instead?

src/service/AnnounceService.ts
```diff
@@ -358,18 +358,39 @@ export class AnnounceService {
...
+                                generationHash,
+                                currency,
+                                transactionService,
+                                listener,
+                                ready,
+                                nodeAccount.name,
+                            );
+                        }
                     } else {
                         await this.announceAggregateComplete(
```
rg911, line 395: does the modifyMultisig tx needs to be announced as a bonded transaction? Using transactions.length might not always be safe. What if modifyMultisig is part of the transactions array? what about using array.filter?

src/service/ModifyMultisigService.ts
```diff
@@ -0,0 +1,191 @@
...
+};
+
+const logger: Logger = LoggerFactory.getLogger(LogType.System);
+
+export class ModifyMultisigService implements TransactionFactory {
+    public static readonly defaultParams: ModifyMultisigParams = {
+        target: BootstrapUtils.defaultTargetFolder,
+        useKnownRestGateways: false,
+        ready: false,
+        url: 'http://localhost:3000',
```
rg911, line 52: default to hardcoded http?

src/service/AnnounceService.ts
yilmazbahadir: why do you think so?

rg911: I mean you can use instanceof, but the SDK has already got a transactionType enum. could that be neat using the type?

src/service/ModifyMultisigService.ts
yilmazbahadir: yes and default to localhost :)
Copied from: https://github.com/symbol/symbol-bootstrap/blob/56239efa3100116ea294c570ddab317160e0e00b/src/service/LinkService.ts#L77
I think it's fine for this PR, the node list loading logic will be changed after @fboucquez 's (soon to be raised) PR.

src/commands/modifyMultisig.ts
```diff
@@ -0,0 +1,67 @@
...
+            char: 'r',
+        }),
+        minApprovalDelta: flags.integer({
+            description:
+                'Delta of signatures needed to approve a transaction. ' +
+                '0 means no change, a positive(+) number means increment and a negative(-) number means decrement to the actual value.',
+            char: 'a',
+        }),
+        addressAdditions: flags.string({
+            description: 'Cosignatory accounts addresses to be added (separated by a comma).',
```
fboucquez, line 46: have you tried `multiple: true`, cli handles lists of strings. Does it work nicely?

src/service/ModifyMultisigService.ts
fboucquez: Yes, it's fine, default url is when you are running the link/modify command on the same box that you have launched the node. Most people would either provide an `--url` or use `--useKnownRestGateways`

fboucquez (changes requested):
Please also run `npm run style:fix` to update the index.ts and docs

I have done the following testing:

Multisig M, cosigners A,B,C , ServiceOperatorAccount S. It looks really nice:

1) Convert to multisig M using M private key. It creates a bonded where All A,B,C need to cosign.
2) Modify mutlisg M using A cosigner private key. It creates a bonded where some of B,C to cosign
3) Convert to multisig  M using S private key. It creates a bonded where all M, A, B, C need to cosign (wallet warning shown as expected)
4) Modify multisig M using S private key. It creates a bonded where some A, B, C need to cosign (wallet warning shown as expected)

Could the command ask (optionally) for cosigners private keys. Use case, I'm a regular user that wants to convert my main account to multisig for security reasons. I could provide my cosigners private keys on the spot without bonded, using full aggregate complete. This could be an improvement though, the service providers should never know any cosigner private key.

src/commands/modifyMultisig.ts
yilmazbahadir: Yes, tried multiple addresses separated by comma, works fine 👍

yilmazbahadir:
> Please also run `npm run style:fix` to update the index.ts and docs
> 
> I have done the following testing:
> 
> Multisig M, cosigners A,B,C , ServiceOperatorAccount S. It looks really nice:
> 
> 1. Convert to multisig M using M private key. It creates a bonded where All A,B,C need to cosign.
> 2. Modify mutlisg M using A cosigner private key. It creates a bonded where some of B,C to cosign
> 3. Convert to multisig  M using S private key. It creates a bonded where all M, A, B, C need to cosign (wallet warning shown as expected)
> 4. Modify multisig M using S private key. It creates a bonded where some A, B, C need to cosign (wallet warning shown as expected)
> 
> Could the command ask (optionally) for cosigners private keys. Use case, I'm a regular user that wants to convert my main account to multisig for security reasons. I could provide my cosigners private keys on the spot without bonded, using full aggregate complete. This could be an improvement though, the service providers should never know any cosigner private key.

Great! Will run the `style:fix` script. Let's implement this improvement with a different PR, created https://github.com/symbol/symbol-bootstrap/issues/308.
Currently writing some unit tests and also adding some parameter sanity checks, will be updating the PR shortly.

src/commands/modifyMultisig.ts
fboucquez: I mean, oclif `multiple: true` field in the flags.string. It users cli style multiple values

fboucquez: ignore, --addressAdditions A,B,C its easier than --addressAdditions A --addressAdditions B --addressAdditions C

does  `--addressAdditions` (empty) work? empty list no prompt.

yilmazbahadir: I think so
>ignore, --addressAdditions A,B,C is easier than --addressAdditions A --addressAdditions B --addressAdditions C

No it doesn't, it gives => No Error: Flag --addressAdditions expects a value
> does --addressAdditions (empty) work? empty list no prompt.

If we need to support no-prompt command as well, I think we have 2 options here:
1. Remove all the prompts, use only the flag values
2. Add the following flags --no-addressAdditions, --no-addressDeletions 
    and the command will look like
    `symbol-bootstrap modifyMultisig --no-addressAdditions --no-addressDeletions --minApprovalDelta=1 --minRemovalDelta=0`
I'd probably vote for 2.

yilmazbahadir: Updated the code as it will not prompt for the following.
`--addressAdditions='' --addressDeletions='' `

PR #309, opened 2021-10-31
fix: removed hardcoded node list from preset. Using statistic services

fboucquez:


src/service/RemoteNodeService.ts
```diff
@@ -113,4 +123,105 @@ export class RemoteNodeService {
             .filter((i) => i)
             .map((i) => i as RepositoryInfo);
     }
+
+    public async getRestUrls(presetData: ConfigPreset): Promise<string[]> {
```
yilmazbahadir, line 127: There are multiple methods taking `presetData` as an argument. Since all of the methods are instance methods, might be better to remove it from method arguments and move it to the constructor instead and access it with `this.presetData`.

yilmazbahadir:
Great work mate! Just a small comment and the following.

Can you rebase your branch on top of `dev`? Looks like there is a config, the following code will probably need a refactor and might move into `TransactionUtils` (also used by `ModifyMultisigService`) .
https://github.com/symbol/symbol-bootstrap/blob/c846846245e02d8886fac5985374da0a15b05a83/src/service/AnnounceService.ts#L121

yilmazbahadir (approved):

PR #315, opened 2021-11-10
fix: Fixed `symbol-statistics-service-typescript-fetch-client` exception handling.

fboucquez:
Fixes https://github.com/symbol/symbol-bootstrap/issues/314

The issue wasn't the proxy, the issue was how the generated `symbol-statistics-service-typescript-fetch-client` handled errors. I have improved that library so the exceptions are handled internally. 

The plan is to move the rest-client generator to a submodule of  https://github.com/symbol/statistics-service that automatically builds clients when updating openapi.

### DIFF
--- a/test/service/RemoteNodeService.test.ts
+++ b/test/service/RemoteNodeService.test.ts
@@ -553,4 +553,32 @@ describe('RemoteNodeService', () => {
             },
         ]);
     });
+
+    it('getPeerInfos unknown statisticsServiceUrl', async () => {
+        presetData.statisticsServiceUrl = 'https://testnet.symbol.invalid';
+        const service = new RemoteNodeService(presetData, false);
+        const peerInfos = await service.getPeerInfos();
+        // only static nodes are returned when the statistics service client fails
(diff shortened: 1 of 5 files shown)


test/service/RemoteNodeService.test.ts
```diff
@@ -553,4 +553,32 @@ describe('RemoteNodeService', () => {
...
+            {
+                publicKey: 'AAAAE7EAEEAE61EF0C50B4D05931F4325F69081B1B074D31E094C4B21E8CFB3D',
+                endpoint: { host: 'someStaticPeer', port: 7900 },
+                metadata: { name: 'someStaticPeer', roles: 'Peer,Api' },
+            },
+        ]);
+    });
+
+    it('getPeerInfos invalid statisticsServiceUrl path', async () => {
+        presetData.statisticsServiceUrl = 'https://testnet.symbol.services/invalid';
```
Jaguar0625, line 572: you could refactor these two tests since this url value is the only difference

Jaguar0625 (approved):

PR #318, opened 2021-11-12
Fix peers json not getting generated correctly.

Wayonb:
The peers json files does not get generated correctly after the statistics service implementation.

The issues are
1. API nodes are been placed in peers-p2p.json
2. Nodes that are not API been put into the peers-api.json

For local network, peer-0 and peer-1 nodes were not able to discover each other.
For mainnet, this would most likely cause the peers-api.json to be empty due to the fact that most of the network is dual nodes(includes api).


fboucquez (approved):
Good finding!

PR #321, opened 2021-11-15
fix: Removed npm module root folder location param.

fboucquez:
It can be resolved without client code intervention, it works better when bootstrap is installed as a library for an app.

Another pr for better custom app integration

To be merged after 1.1.1 release

### DIFF
--- a/src/commands/compose.ts
+++ b/src/commands/compose.ts
@@ -47,6 +47,6 @@ export default class Compose extends Command {
             CommandUtils.passwordPromptDefaultMessage,
             true,
         );
-        await new BootstrapService(this.config.root).compose(flags);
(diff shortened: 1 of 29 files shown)


fboucquez:
> are you sure that no one using bootstrap as a library would want to use a custom path?

The only reason I can think:
A dev using bootstrap as a lib and wants to customize anything on the `config` or `presets` folder (that cannot be done via custom presets). If the dev does that, it would need to copy and paste both folders in a different location. 

Should we keep dev provided root folder? If a dev needs to do something like that, it might be better to do a feature request to improve bootstrap customization. I am pretty sure I am the only one using bootstrap as lib in my network tools.

Jaguar0625 (approved):

PR #329, opened 2021-11-24
Added how to style fix and generate commands docs

fboucquez:


segfaultxavi (approved):
Super-minor comment for next time: code blocks should always indicate the language (in this case, `bash`).

Jaguar0625 (approved):

PR #331, opened 2021-11-25
feature: allowing user provided --assembly and --preset

fboucquez:
Continuation of https://github.com/symbol/symbol-bootstrap/pull/328.

This pr allows users to provide their own network preset and seed folder. 

Use cases:

1) We are creating a new testnet network that hasn't been published in bootstrap yet (because it's being created).  To create these nodes, we would do:

`symbol-bootstrap start --preset new-testnet-network.yml --assembly dual`

Once we create the "seed" nodes, we can move `new-testnet-network.yml` and nemesis block into bootstrap official `testnet`. Once bootstrap has the new testnet network.yml and seed, it can be released so people can use `symbol-bootstrap start --preset testnet --assembly dual` to connect to the brand new network.

2) We are creating a new experimental testnet network that will "never" be part of bootstrap, for example,  garush.dev:

`symbol-bootstrap start --preset garush-network.yml --assembly dual`

In this case, we may or may not share the network.yml and nemesis block with the community, depending on if we want community nodes to be connected or not.

In addition, nemesis information is included in all the network.yml files. This allows Symbol-Network to 'clone' an existing configuration that then the network admin can tune, for example, changing the name of the coin, the supply, divisibility, etc.  The garush.dev network is a clone of testnet where I changed the network namespace and coin name


src/commands/config.ts
```diff
@@ -33,15 +35,17 @@ export default class Config extends Command {
...
         password: CommandUtils.passwordFlag,
         noPassword: CommandUtils.noPasswordFlag,
-        preset: flags.enum({
+        preset: flags.string({
             char: 'p',
-            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file.`,
-            options: Object.keys(Preset).map((v) => v as Preset),
+            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file. Options are: ${Object.keys(
+                Assembly,
+            ).join(', ')}, my-custom-network.yml (advance)`,
```
Jaguar0625, line 42: advance => advanced ?

src/commands/wizard.ts
```diff
@@ -49,8 +41,16 @@ export enum Network {
...
     testnet = 'testnet',
     localNetwork = 'localNetwork',
+    customNetwork = 'customNetwork',
 }
 
+export const assemblies: Record<Network, Assembly[]> = {
+    [Network.mainnet]: [Assembly.dual, Assembly.peer, Assembly.api],
+    [Network.testnet]: [Assembly.dual, Assembly.peer, Assembly.api, Assembly.demo],
+    [Network.localNetwork]: [Assembly.multinode, Assembly.dual, Assembly.peer, Assembly.api, Assembly.demo],
+    [Network.customNetwork]: [Assembly.dual, Assembly.peer, Assembly.api],
```
Jaguar0625, line 51: not sure why we need mainnet|testnet|customNetwork, what is _not_ stored in assembly/preset files at this point. 

additionally, should probably use consistent naming mainnet/testnet/local(net?)/custom(net?)

Jaguar0625 (changes requested):

src/commands/wizard.ts
fboucquez: `customNetwork` is when a third-party user wants to connect to a custom network (created with symbol-network). The network admin should have shared the custom network.yml preset and nemesis-seed folder. This is just adding the option on the wizard.

About the naming, I'm keen to split the local network's bootstrap preset into -p singleCurrency and -p dualCurrency (current -p bootstrap). What do you think? 

`symbol-bootstrap -p singleCurrency` would be an agnostic local network with just one currency.

Jaguar0625: i'd probably drop testnet preset too since we expect multiple testnets in future it might be confusing

fboucquez: Related to the https://github.com/symbol/symbol-bootstrap/pull/341 discussion. 
I would prefer to keep joeynet or any other "commonly" used testnet to simplify the user experience. Other uncommon or experimental testnet can be shared by providing the custom network.yml and seed folder, this PR allows it.

Jaguar0625: we can discuss that more in discord

Jaguar0625 (approved):

PR #337, opened 2021-11-28
Nemesis configuration improvements

fboucquez:
~1. Fixed config/start --preset and --assembly options (it was the other way around)~ Already merged in another pr
2. Removed unnecessary currency names from custom preset. It can be provided and resolved from the `nemesis.mosaics` field
3. Moved "opt-in" addresses/balances from `nemesis` to `nemesis.mosiac`. User (or symbol-network) can provide balances for multiple mosaics, although most of the time there is only one nemesis mosaic (like xym).
4. Allow a node to be excluded from nemesis (no balance or linked transactions). 
5. Allow users to provide node nemesis balance. If the balance is calculated, it will be less than  `maxHarvesterBalance` 
6. Better nemesis supply distribution and exception handling. Added unit tests.  

If this is too much I can split further. Note I would be creating a new PR where the genesis related configuration is moved to their own class/service

### DIFF
--- a/src/model/ConfigPreset.ts
+++ b/src/model/ConfigPreset.ts
@@ -33,6 +33,11 @@ export interface DockerServicePreset {
     dockerComposeDebugMode?: boolean;
 }
 
+export interface CurrencyDistribution {
+    address: string;
+    amount: number;
+}
+
 export interface MosaicPreset {
     name: string;
     repeat?: number;
@@ -44,8 +49,9 @@ export interface MosaicPreset {
     isTransferable: boolean;
     isSupplyMutable: boolean;
     isRestrictable: boolean;
-    accounts: number;
-    currencyDistributions: { address: string; amount: number }[];
+    // options are generate x random accounts or provide their accounts public keys.
+    accounts: number | string[];
+    currencyDistributions: CurrencyDistribution[];
 }
 
 export interface DatabasePreset extends DockerServicePreset {
@@ -57,7 +63,6 @@ export interface DatabasePreset extends DockerServicePreset {
 export interface NemesisPreset {
     binDirectory: string;
     mosaics: MosaicPreset[];
-    balances?: Record<string, number>;
     transactions?: Record<string, string>;
     nemesisSignerPrivateKey: string;
     transactionsDirectory: string;
@@ -259,6 +264,7 @@ export interface NodePreset extends DockerServicePreset, Partial<NodeConfigPrese
     host?: string;
     roles?: string;
     friendlyName?: string;
+    excludeFromNemesis: boolean;
 
     // Optional private keys. If not provided, bootstrap will generate random ones.
     mainPrivateKey?: string;
@@ -272,6 +278,8 @@ export interface NodePreset extends DockerServicePreset, Partial<NodeConfigPrese
 
     vrfPrivateKey?: string;
     vrfPublicKey?: string;
+
+    balances?: number[];
 
     //Broker specific
     brokerName?: string;
@@ -416,9 +424,6 @@ export interface CommonConfigPreset extends NodeConfigPreset, GatewayConfigPrese
     currencyMosaicId: string;
     harvestingMosaicId: string;
     baseNamespace: string;
-    // This next 2 would be removed in a later merge.
-    currencyName?: string;
-    harvestingName?: string;
     networkType: NetworkType;
     votingKeyDesiredLifetime: number;
     votingKeyDesiredFutureLifetime: number; // How in the future voting key files need to be generated. By default, 1 months before expiring..
--- a/src/service/ConfigLoader.ts
+++ b/src/service/ConfigLoader.ts
@@ -156,7 +156,7 @@ export class ConfigLoader {
         const networkType = presetData.networkType;
         const mosaics: MosaicAccounts[] = [];
         presetData.nemesis.mosaics.forEach((m, mosaicIndex) => {
-            const accounts = this.generateAddresses(networkType, privateKeySecurityMode, m.accounts || 1);
+            const accounts = this.generateAddresses(networkType, privateKeySecurityMode, m.accounts);
             const id = MosaicId.createFromNonce(MosaicNonce.createFromNumber(mosaicIndex), nemesisSignerAddress).toHex();
             mosaics.push({
                 id: id,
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -186,6 +185,12 @@ Although some combinations can be done, they may not be really useful. Examples
 -   `$ symbol-bootstrap start -p mainnet -a multinode`
 -   `$ symbol-bootstrap start -p testnet -a multinode`
 
+A custom network preset file can also be provided. This is useful when you have your own custom Symbol network, and you want other nodes to join.
+For this case, you provide your own `networkPreset.yml` and nemesis feed folder. The node admin can then run:
```
segfaultxavi, line 189:
```suggestion
For this case, you provide your own `networkPreset.yml` and nemesis seed folder. The node admin can then run:
```

src/commands/config.ts
```diff
@@ -33,15 +35,17 @@ export default class Config extends Command {
         target: CommandUtils.targetFlag,
         password: CommandUtils.passwordFlag,
         noPassword: CommandUtils.noPasswordFlag,
-        preset: flags.enum({
+        preset: flags.string({
             char: 'p',
-            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file.`,
-            options: Object.keys(Preset).map((v) => v as Preset),
+            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file. Options are: ${Object.keys(
+                Assembly,
```
segfaultxavi, line 41:
```suggestion
                Preset,
```

src/commands/config.ts
```diff
@@ -33,15 +35,17 @@ export default class Config extends Command {
...
+            ).join(', ')}, my-custom-network.yml (advance)`,
         }),
-        assembly: flags.enum({
+        assembly: flags.string({
             char: 'a',
-            description: `The assembly that defines the node(s) layout. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file.`,
-            options: Object.keys(Assembly).map((v) => v as Assembly),
+            description: `The assembly that defines the node(s) layout. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file. Options are: ${Object.keys(
+                Preset,
```
segfaultxavi, line 47:
```suggestion
                Assembly,
```

src/commands/config.ts
```diff
@@ -33,15 +35,17 @@ export default class Config extends Command {
...
         password: CommandUtils.passwordFlag,
         noPassword: CommandUtils.noPasswordFlag,
-        preset: flags.enum({
+        preset: flags.string({
             char: 'p',
-            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file.`,
-            options: Object.keys(Preset).map((v) => v as Preset),
+            description: `The network preset. It can be provided via custom preset or cli parameter. If not provided, the value is resolved from the target/preset.yml file. Options are: ${Object.keys(
+                Assembly,
+            ).join(', ')}, my-custom-network.yml (advance)`,
```
segfaultxavi, line 42: What does `(advance)` mean?

segfaultxavi (changes requested):
Some mixed-up options...

src/commands/config.ts
fboucquez: Thanks!

src/model/ConfigPreset.ts
```diff
@@ -259,6 +264,7 @@ export interface NodePreset extends DockerServicePreset, Partial<NodeConfigPrese
     host?: string;
     roles?: string;
     friendlyName?: string;
+    excludeFromNemesis: boolean;
```
Jaguar0625, line 267: why is this option needed?

src/service/ConfigLoader.ts
```diff
@@ -130,43 +130,107 @@ export class ConfigLoader {
...
+                return d.amount;
+            })
+            .reduce((a, b) => a + b, 0);
+    }
+    private processNemesisBalances(presetData: ConfigPreset, addresses: Addresses, nemesisSignerAddress: Address): MosaicAccounts[] {
+        const privateKeySecurityMode = CryptoUtils.getPrivateKeySecurityMode(presetData.privateKeySecurityMode);
+        const networkType = presetData.networkType;
+        const mosaics: MosaicAccounts[] = [];
+        presetData.nemesis.mosaics.forEach((m, mosaicIndex) => {
+            const accounts = this.generateAddresses(networkType, privateKeySecurityMode, m.accounts || 1);
```
Jaguar0625, line 159: why || 1 ?

src/model/ConfigPreset.ts
fboucquez: One use case,

When symbol-networks creates the 'genesis block' node (creating the network nemesis block), all the balances and link transactions are explicitly provided by the tool (according to the network layout defined in a previous step). I don't want the nemesis block node to be included at all in the nemesis block.

fboucquez: This step: https://hackmd.io/Ak7Sl1A6SkKV7ga0QW00LA#Nemesis-generation

src/service/ConfigLoader.ts
fboucquez: Removed and unit tested

src/model/ConfigPreset.ts
Jaguar0625: what is this use case? where did this requirement come from?

fboucquez: It's a helper ffeature for my symbol-network tool

Wayonb: I see the ``excludeFromNemesis`` property getting use [here.](https://github.com/symbol/symbol-testnet/blob/deb438f6bf5114136af6b89b573c59f2f70d6e33/joeynet/nemesis-target/preset.yml#L775) 
Is this just to create local nodes?

fboucquez: Yes, it's used by the tool in one of the steps. The step is when the local "nemesis" node is created. That nemesis node generates the nemesis block where all the other final node balances and transactions are provided. I don't want to include anything to that node in the nemesis block so the exclude flag.

Jaguar0625: are you saying that there's a local node that is generating all the other nodes?

fboucquez: The local node's nemesis block (seed folder) is copied over the other nodes when they are generated.

Jaguar0625 (approved):

PR #338, opened 2021-11-28
fix: voting key file provider

fboucquez:
This improvement allows devs to use another mechanism for voting key file creating. I would be using this in 2 ways:

1) I have a suite of tests to be pushed that does deep-dir comparison of different setups. The unit test needs a way to generate static files so we can do the assertions on all the files.  

The incoming test and unit test file provider:
https://github.com/symbol/symbol-bootstrap/blob/networkPresetFile/test/service/BootstrapService.test.ts
https://github.com/symbol/symbol-bootstrap/blob/networkPresetFile/test/service/BootstrapService.test.ts#L55

I need other PRs before pushing those unit tests.

2) Symbol services pre-create the voting files when generating the nemesis block on new networks, so they can then be used when creating each node. Those files are stored in the symbol-network's key-store.yml.

### DIFF
--- a/src/service/VotingKeyFileProvider.ts
+++ b/src/service/VotingKeyFileProvider.ts
@@ -19,7 +19,7 @@ import { Account } from 'symbol-sdk';
 import { Logger } from '../logger';
 import { ConfigPreset, NodeAccount, NodePreset } from '../model';
 import { BootstrapUtils } from './BootstrapUtils';
-import { VotingKeyFile, VotingUtils } from './VotingUtils';
+import { VotingUtils } from './VotingUtils';
 
 export interface VotingKeyParams {
     presetData: ConfigPreset;
@@ -31,8 +31,12 @@ export interface VotingKeyParams {
     votingKeyEndEpoch: number;
 }
 
+export interface VotingKeyCreationResult {
+    publicKey: string;
+}
+
 export interface VotingKeyFileProvider {
-    createVotingFile(params: VotingKeyParams): Promise<VotingKeyFile>;
+    createVotingFile(params: VotingKeyParams): Promise<VotingKeyCreationResult>;
 }
 
 export class NativeVotingKeyFileProvider implements VotingKeyFileProvider {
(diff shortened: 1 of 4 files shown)


src/service/VotingKeyFileProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        const votingPrivateKey = votingAccount.privateKey;
+        const votingUtils = new VotingUtils();
+        this.logger.info('Voting file is created using the native typescript voting key file generator!');
+        const votingFile = await votingUtils.createVotingFile(votingPrivateKey, votingKeyStartEpoch, votingKeyEndEpoch);
+        writeFileSync(join(votingKeysFolder, privateKeyTreeFileName), votingFile);
+        return {
+            publicKey: votingAccount.publicKey,
+            startEpoch: votingKeyStartEpoch,
+            endEpoch: votingKeyEndEpoch,
+            filename: privateKeyTreeFileName,
```
Jaguar0625, line 57: not sure why you're forwarding these three parameters? since the caller is passing them in, the caller will have these already.

src/service/VotingKeyFileProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+    constructor(private readonly logger: Logger, private readonly user: string) {}
+    public async createVotingFile({
+        presetData,
+        votingKeysFolder,
+        privateKeyTreeFileName,
+        votingKeyStartEpoch,
+        votingKeyEndEpoch,
+    }: VotingKeyParams): Promise<VotingKeyFile> {
+        this.logger.info(`Voting file is created using docker and catapult.tools.votingkey`);
+        const votingAccount = Account.generateNewAccount(presetData.networkType);
```
Jaguar0625, line 72: seems like this could potentially be passed in as well, since all implementations will need to create a voting key

gimre-xymcity (dismissed):

src/service/VotingKeyFileProvider.ts
fboucquez: The caller is calculating the start and end epoch of the voting key file to be created. Then the implementation is creating the key and the file with those values.

src/service/VotingKeyFileProvider.ts
fboucquez: The implementation is resolving the private key. Most of the time would be a new random one. 
It's not random for the 2 use cases explained above. Unit tests for assertions, and when the file was previously created and cached somewhere else.

src/service/VotingKeyFileProvider.ts
Jaguar0625: yes, but why are you _returning_ them ?

fboucquez: Got it! removed return params.

Jaguar0625 (approved):

PR #339, opened 2021-11-29
fix: removed unused files after nemesis block is generated.

fboucquez:
This PR removes the nemesis block temp files after generation. We can then have deterministic nemesis seed folder generation and compare folders in unit tests (Future PR).

### DIFF
--- a/src/service/NemgenService.ts
+++ b/src/service/NemgenService.ts
@@ -49,10 +49,8 @@ export class NemgenService {
             `${presetData.catapultAppFolder}/bin/catapult.tools.nemgen`,
             '--resources=/server-config',
             '--nemesisProperties=./server-config/block-properties-file.properties',
+            '--useTemporaryCacheDatabase',
         ];
-        if (presetData.nemgenUseTemporaryCacheDatabase) {
-            cmd.push(`--useTemporaryCacheDatabase`);
-        }
 
         const binds = [`${serverConfigWorkingDir}:/server-config`, `${nemesisWorkingDir}:/nemesis`];
 
--- a/src/service/NemgenService.ts
+++ b/src/service/NemgenService.ts
@@ -36,7 +36,7 @@ export class NemgenService {
         }
 
         const nemesisWorkingDir = BootstrapUtils.getTargetNemesisFolder(target, true);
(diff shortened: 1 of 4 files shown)


Jaguar0625:
> This PR removes the nemesis block temp files after generation. 

Removes from where? Local machine or a deployed server?

fboucquez:
Hi @Jaguar0625 , the issue description wasn't the best. 

When user runs -p bootstrap, the nemgen creates all these files:

image

This PR is to clean up the files the catapult client doesn't require saving space and helping me with the unit test suite that it's coming

Jaguar0625:
why are you not running nemgen with `useTemporaryCacheDatabase` option?

fboucquez:
> why are you not running nemgen with `useTemporaryCacheDatabase` option?

Thanks! I didn't know the option.

src/service/NemgenService.ts
```diff
@@ -50,24 +50,38 @@ export class NemgenService {
             '--resources=/server-config',
             '--nemesisProperties=./server-config/block-properties-file.properties',
         ];
+        if (presetData.nemgenUseTemporaryCacheDatabase) {
+            cmd.push(`--useTemporaryCacheDatabase`);
+        }
```
Jaguar0625, line 55: why did you add option for this? when would you want to run WITHOUT it?

fboucquez: Only for testing? By providing: `nemgenUseTemporaryCacheDatabase: true` on custom preset.

fboucquez: I have never "read" the cache database, if you think nobody will, I can remove the option and always use temp

Jaguar0625: i think it is always acceptable to use this option. but you can check with @gimre-xymcity.

fboucquez: @gimre-xymcity would somebody ever use the cache database for debugging?

gimre-xymcity: nope

Jaguar0625 (approved):

src/service/NemgenService.ts
fboucquez: Removed the option

PR #340, opened 2021-11-29
fix: removed unused certificate generation files

fboucquez:
It reduces the certificate folder, making it lighter for dir-comparision future unit tests.


Jaguar0625 (changes requested):
you only need these 5 files on the server:
ca.crt.pem
ca.pubkey.pem
node.crt.pem
node.full.crt.pem
node.key.pem

fboucquez:
Thanks @Jaguar0625, less is more. Removing other unused files.

test/service/CertificateService.test.ts
```diff
@@ -70,26 +70,9 @@ describe('CertificateService', () => {
...
-            'new_certs',
-            'node.cnf',
-            'node.crt.pem',
-            'node.csr.pem',
-            'node.full.crt.pem',
-            'node.key.pem',
-            'serial.dat',
-            'serial.dat.old',
-        ]);
+        expect(files).deep.eq(['ca.cert.pem', 'ca.pubkey.pem', 'metadata.yml', 'node.crt.pem', 'node.full.crt.pem', 'node.key.pem']);
```
Jaguar0625, line 73: i assume metadata.yml was kept intentionally?

Jaguar0625 (approved):

test/service/CertificateService.test.ts
fboucquez: Yes, metadata is mine. It holds the used main and node public keys. It's not easy nodejs to read the pem files without a library.

PR #341, opened 2021-11-29
feat: joeynet reset

fboucquez:
This joynet configuration is the one I generated using the symbol-network tool. This pr "officialize" it by distributing it with the bundled testnet reset.

The network.yml configuration would be slightly different once https://github.com/symbol/symbol-bootstrap/pull/331 is merged. 

The network, seed nodes configuration, and accounts can be found in the new private repo https://github.com/symbol/symbol-testnet

This PR includes the initial node seeds. Statistic service would be "empty" until we have launched the seed nodes.  We would need to reset and reconfigure the statistic service soon after joeynet network launch.

We can eventually remove the static list from joeynet network.yml once Statistic service is up to date. 
Bootstrap supports both static and/or dynamic stats service lists.


Jaguar0625 (changes requested):
joeynet is one of (potentially many) testnets. i don't think it makes sense to replace testnet with it. instead, i think we should drop testnet/mainnet entirely and just specify preset for mainnet/joeynet/etc.

i'd also suggest having a separate repo for the seed, so we can have releases for it similar to the release we have for mainnet seed: https://github.com/symbol/catapult-client/releases/download/v1.0.0.0/nemesis-seed.zip.

fboucquez:
Thans @Jaguar0625  for the feedback,

> joeynet is one of (potentially many) testnets. i don't think it makes sense to replace testnet with it. instead, i think we should drop testnet/mainnet entirely and just specify preset for mainnet/joeynet/etc.
> 

I agree, I am working on all the toolings for multiple testnet support. The shared assemblies [pr](https://github.com/symbol/symbol-bootstrap/pull/328), the external network preset yml and seed for [pr](https://github.com/symbol/symbol-bootstrap/pull/331), symbol-network cli, the [blog](https://hackmd.io/Ak7Sl1A6SkKV7ga0QW00LA?view), etc. 

Joeynet would be one of many, some official, some experimental (garsuh.dev), some community, etc.

I have asked if we should keep the -p joeynet preset for this reset but we agreed to keep the -p testnet name.

https://discord.com/channels/856325968096133191/912347099155292163/912398258188484618

I think this is fine, the current NGL testnet is reaching the EOL and we are not maintaining it anymore. It seems like Joeynet is becoming the principal testnet now. The -p testnet name is fine, we don't need to change the documentation around how to run a testnet node. We are also planning to reuse old testnet URLs like faucet, stats service, and explorer. If we keep both, we would need different URLs for Joeynet

> i'd also suggest having a separate repo for the seed, so we can have releases for it similar to the release we have for mainnet seed: https://github.com/symbol/catapult-client/releases/download/v1.0.0.0/nemesis-seed.zip.

I like having the most common official network seed and network.yml bundled in bootstrap. We don't need to maintain a separate hosting for the seeds and it's 'offline' friendly when generating the node configuration.

fboucquez:
@Wayonb @Jaguar0625 , could we merge this one? it's blocking me for the testnet reset. Once I have this one I can create the new zips

Jaguar0625 (approved):

Wayonb:
Create this issue to move the seed out of the bootstrap - https://github.com/symbol/symbol-bootstrap/issues/342

PR #345, opened 2021-12-06
feat: Allowing user provided compose services via custom preset

fboucquez:
This small feature allows users to tune the generated compose, for example adding a logger or third part compose service.

I would use this feature to inject AWS cloudwatch logging as Ravi requested.


Jaguar0625:
i still don't think this feature makes sense in bootstrap since the changes can be easily done as a post-build step. @gimre-xymcity and @0x6861746366574 can comment though.

fboucquez:
> i still don't think this feature makes sense in bootstrap since the changes can be easily done as a post-build step. @gimre-xymcity and @0x6861746366574 can comment though.

It's an optional feature to make it easier to extend compose. Post-build could be done but it would be harder for the end-user to (or symbol-network) to parse and patch docker composes when it's created the first time and every time is --upgraded. I would definitely use an extension feature like this. The customization stays on the custom-preset.yml

A custom user post-build would modify the target folder which I prefer not to. 

I have updated the test custom preset to add logging to each service.

PR #346, opened 2021-12-10
fix: Created RuntimeService from BootstrapUtils

fboucquez:
Created RuntimeService from BootstrapUtils. One PR to start splitting the classes.


src/service/Utils.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+
+    public static validateIsTrue(value: boolean, message: string): void {
+        if (!value) {
+            throw new Error(message);
+        }
+    }
+    public static logSameLineMessage(message: string): void {
+        process.stdout.write(RuntimeService.isWindows() ? '\x1b[0G' : '\r');
```
Jaguar0625, line 24: dependency on RuntimeService seems unnatural. probably better to move isWindows into this class.

test/service/RuntimeService.test.ts
```diff
@@ -0,0 +1,62 @@
+import { expect } from 'chai';
+import { it } from 'mocha';
+import { LoggerFactory, LogType, RuntimeService } from '../../src';
+const logger = LoggerFactory.getLogger(LogType.Silent);
+const service = new RuntimeService(logger);
+describe('RuntimeService', async () => {
+    it('exec when valid', async () => {
```
Jaguar0625, line 7: does this test pass on windows?

Jaguar0625 (changes requested):

src/service/Utils.ts
fboucquez: Created OSUtils

test/service/RuntimeService.test.ts
Jaguar0625: cc @gimre-xymcity

fboucquez: Jumped over to windows and now all tests are running on windows. Afair,  Travis's windows vm doesn't support docker. Keen to migrate to github actions if have the time (maybe actions works better, need to revisit)

PR #350, opened 2021-12-14
fix: Account Resolver

fboucquez:
fix: ConfigLoader split -> MigrationService, AddressesService, ConfigurationUtils
fix: addresses.sinkAddress is not generated when it's not required

Note: I had to fix some imports. If not, oclif errors when generating the docs and the cli metadata. I think it's a circular dependency somewhere.

### DIFF
--- a/src/service/AccountResolver.ts
+++ b/src/service/AccountResolver.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2021 NEM
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Account, NetworkType } from 'symbol-sdk';
+import { CertificatePair, KeyName } from '../';
+
+/**
+ * Delegate that knows how to retrieve or generate accounts.
+ *
+ * Implementations of this interface could for example prompt or load accounts from a key store.
+ */
+export interface AccountResolver {
+    resolveAccount(
+        networkType: NetworkType,
+        account: CertificatePair | undefined,
+        keyName: KeyName,
+        nodeName: string | undefined,
+        operationDescription: string,
+        generateErrorMessage: string | undefined,
+    ): Promise<Account>;
+}
+
+/**
+ * Basic no prompt implementation. If the account cannot be resolved, it won't be prompted.
+ */
+export class DefaultAccountResolver implements AccountResolver {
+    async resolveAccount(
+        networkType: NetworkType,
+        account: CertificatePair | undefined,
+        keyName: KeyName,
+        nodeName: string,
+        operationDescription: string,
+        generateErrorMessage: string | undefined,
+    ): Promise<Account> {
+        if (!account) {
+            if (generateErrorMessage) {
+                throw new Error(generateErrorMessage);
+            }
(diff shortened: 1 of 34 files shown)


src/service/AnnounceService.ts
```diff
@@ -225,16 +226,13 @@ export class AnnounceService {
...
-                        nodeAccount.name,
-                        'signing a transaction',
-                    ),
+                return this.accountResolver.resolveAccount(
                     networkType,
+                    nodeAccount.main,
+                    KeyName.Main,
+                    nodeAccount.name,
+                    'signing a transaction',
+                    'Should not generate!',
```
gimre-xymcity, line 235: that error does not seem very descriptive

src/service/BootstrapAccountResolver.ts
```diff
@@ -0,0 +1,76 @@
...
+        operationDescription: string,
+        generateErrorMessage: string | undefined,
+    ): Promise<Account> {
+        if (!account) {
+            if (generateErrorMessage) {
+                throw new KnownError(generateErrorMessage);
+            }
+            this.logger.info(`Generating ${keyName} account...`);
+            return Account.generateNewAccount(networkType);
+        }
```
gimre-xymcity, line 40: this seems like exact part from DefaultAccountResolver, refactor?

src/service/BootstrapAccountResolver.ts
```diff
@@ -0,0 +1,76 @@
...
+                } else {
+                    const enteredAccount = Account.createFromPrivateKey(privateKey, networkType);
+                    if (enteredAccount.publicKey.toUpperCase() !== account.publicKey.toUpperCase()) {
+                        this.logger.info(
+                            `Invalid private key. Expected address is ${address} but you provided the private key for address ${enteredAccount.address.plain()}.\n`,
+                        );
+                        this.logger.info(`Please re-enter private key.`);
+                    } else {
+                        account.privateKey = privateKey;
+                        return Account.createFromPrivateKey(privateKey, networkType);
```
gimre-xymcity, line 69: return enteredAccount (?)

src/service/BootstrapAccountResolver.ts
```diff
@@ -0,0 +1,76 @@
...
+                    this.logger.info('Please provide the private key.');
+                } else {
+                    const enteredAccount = Account.createFromPrivateKey(privateKey, networkType);
+                    if (enteredAccount.publicKey.toUpperCase() !== account.publicKey.toUpperCase()) {
+                        this.logger.info(
+                            `Invalid private key. Expected address is ${address} but you provided the private key for address ${enteredAccount.address.plain()}.\n`,
+                        );
+                        this.logger.info(`Please re-enter private key.`);
+                    } else {
+                        account.privateKey = privateKey;
```
gimre-xymcity, line 68: is setting this deliberate?

src/service/AccountResolver.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Basic no prompt implementation. If the account cannot be resolved, it won't be prompted.
+ */
+export class DefaultAccountResolver implements AccountResolver {
+    async resolveAccount(
+        networkType: NetworkType,
+        account: CertificatePair | undefined,
+        keyName: KeyName,
+        nodeName: string,
+        operationDescription: string,
+        generateErrorMessage: string | undefined,
```
gimre-xymcity, line 45: is there actually a case (besides tests), where you are NOT providing error message?

(same goes for BootstrapAccountResolver, as theyy implement same interface)

gimre-xymcity (commented):
part 1, more tomorrow

fboucquez:
> part 1, more tomorrow

Thanks gimre for the review. I'm trying to allow libs to provide their own accounts cached somewhere else, on-demand, when bootstrap requires it. I'm using this feature in symbol-network, when the keys are pre-created to generate the nemesis block. The feature is working, but the implementation is not great yet.

src/service/BootstrapAccountResolver.ts
fboucquez: It's to avoid re-prompting for the private key. It's a bit of a secondary effect but it helps.

PR #351, opened 2021-12-17
Verify service improvements.

fboucquez:
It helps test the current bootstrap installation (for symbol-network)
It allows to reuse the reporting and verification errors.
Fixes https://github.com/symbol/symbol-bootstrap/issues/274

### DIFF
--- a/src/service/VerifyService.ts
+++ b/src/service/VerifyService.ts
@@ -32,6 +32,7 @@ export interface ExpectedVersions {
     node: string;
     docker: string;
     dockerCompose: string;
+    symbolBootstrap?: string;
 }
 
 const defaultExpectedVersions: ExpectedVersions = {
@@ -51,15 +52,38 @@ export class VerifyService {
     public async createReport(): Promise<VerifyReport> {
         const lines: ReportLine[] = [];
         const platform = `${os.type()} - ${os.release()} - ${os.platform()}`;
-        lines.push(await this.testNodeJs());
-        const docker = await this.testDocker();
+        lines.push(await this.testNodeJs(this.expectedVersions.node));
+        const docker = await this.testDocker(this.expectedVersions.docker);
         lines.push(docker);
-        lines.push(await this.testDockerCompose());
+        lines.push(await this.testDockerCompose(this.expectedVersions.dockerCompose));
+        if (this.expectedVersions.symbolBootstrap) lines.push(await this.testSymbolBootstrap(this.expectedVersions.symbolBootstrap));
         if (!docker.recommendation) lines.push(await this.testDockerRun());
         if (!BootstrapUtils.isWindows()) {
             lines.push(await this.testSudo());
         }
+
         return { lines, platform };
+    }
+
+    public logReport(report: VerifyReport): void {
+        this.logger.info(`OS: ${report.platform}`);
+        report.lines.forEach((line) => {
+            if (line.recommendation) {
+                this.logger.error(`${line.header}  - Error! - ${line.message} - ${line.recommendation}`);
+            } else {
+                this.logger.info(`${line.header} - OK! - ${line.message}`);
+            }
+        });
+    }
+
+    public validateReport(report: VerifyReport): void {
+        const errors = report.lines.filter((r) => r.recommendation);
+        if (errors.length) {
+            throw new Error(
+                'There has been an error. Check the report: \n' +
+                    errors.map((line) => ` - ${line.header}  - Error! - ${line.message} - ${line.recommendation}`).join('\n'),
+            );
+        }
     }
 
     public loadVersion(text: string): string | undefined {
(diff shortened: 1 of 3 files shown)


gimre-xymcity (approved):

test/service/VerifyService.test.ts
```diff
@@ -104,5 +106,12 @@ describe('VerifyService', () => {
             platform: `${os.type()} - ${os.release()} - ${os.platform()}`,
         };
         expect(report).to.be.deep.eq(expected);
+        service.logReport(report);
+        expect(() => service.validateReport(report)).to.throw(
+            `There has been an error. Check the report:
+ - NodeVersion  - Error! - ${currentNodeJsVersion} - At least version ${expectedVersions.node} is required. Currently installed version is ${currentNodeJsVersion}. Check https://nodejs.org/en/download/package-manager/
+ - Docker Version  - Error! - ${currentDockerVersion} - At least version ${expectedVersions.docker} is required. Currently installed version is ${currentDockerVersion}. Check https://docs.docker.com/get-docker/
+ - Docker Compose Version  - Error! - ${currentDockerComposeVersion} - At least version ${expectedVersions.dockerCompose} is required. Currently installed version is ${currentDockerComposeVersion}. Check https://docs.docker.com/compose/install/`,
```
Jaguar0625, line 114: you seem to be missing test for new (symbolBootstrap) check that was added

src/service/VerifyService.ts
```diff
@@ -74,34 +98,31 @@ export class VerifyService {
...
-        return this.verifyInstalledApp(
-            async () => await this.loadVersionFromCommand(command),
-            header,
-            this.expectedVersions.docker,
-            recommendationUrl,
-        );
+        return this.verifyInstalledApp(() => this.loadVersionFromCommand(command), header, expectedVersion, recommendationUrl);
     }
-    public async testDockerCompose(): Promise<ReportLine> {
+    public async testDockerCompose(expectedVersion: string): Promise<ReportLine> {
```
Jaguar0625, line 114: newline after brace

test/service/VerifyService.test.ts
fboucquez: I didn't fully like having a bootstrap check on bootstrap. it's like checking their own tool. 

The objective of that check is for `symbol-network verify` to check the installed symbol-bootstrap is the correct one. I have refactored this so the bootstrap check (or any other check) can be added with custom actions (symbol-network will add the custom bootstrap check).

src/service/VerifyService.ts
```diff
@@ -41,91 +40,100 @@ const defaultExpectedVersions: ExpectedVersions = {
...
+        return this.loadVersion((await BootstrapUtils.exec(this.logger, command)).stdout.trim());
     }
 
-    public async testDocker(expectedVersion: string): Promise<ReportLine> {
-        const header = 'Docker Version';
-        const command = 'docker --version';
-        const recommendationUrl = `https://docs.docker.com/get-docker/`;
-        return this.verifyInstalledApp(() => this.loadVersionFromCommand(command), header, expectedVersion, recommendationUrl);
+    public async verifyInstalledApp(
+        versionLoader: () => Promise<string | undefined>,
```
Jaguar0625, line 68: when can this return undefined? 
in code below, it looks like it throws on failure.

src/service/VerifyService.ts
```diff
@@ -41,91 +40,100 @@ const defaultExpectedVersions: ExpectedVersions = {
...
+        minVersion: string,
+        recommendationUrl: string,
+    ): Promise<ReportLine> {
+        try {
+            const version = await versionLoader();
+            if (!version) {
+                return {
+                    header,
+                    message: `Version could not be found! Output: ${versionLoader}`,
+                    recommendation: `At least version ${minVersion} is required. Check ${recommendationUrl}`,
```
Jaguar0625, line 79: `At least version ${minVersion} is required. Check ${recommendationUrl}` is used in all recommendations, except in middle it has a prefix with current version status.

src/service/VerifyService.ts
```diff
@@ -41,91 +40,100 @@ const defaultExpectedVersions: ExpectedVersions = {
...
-        const errors = report.lines.filter((r) => r.recommendation);
-        if (errors.length) {
-            throw new Error(
-                'There has been an error. Check the report: \n' +
-                    errors.map((line) => ` - ${line.header}  - Error! - ${line.message} - ${line.recommendation}`).join('\n'),
-            );
-        }
-    }
+export interface VerifyAction {
+    shouldRun(lines: ReportLine[]): Promise<boolean>;
```
Jaguar0625, line 44: why is this returning a promise?

src/service/VerifyService.ts
```diff
@@ -149,8 +157,14 @@ export class VerifyService {
             };
         }
     }
+    shouldRun(lines: ReportLine[]): Promise<boolean> {
+        const dockerIsFine = !!lines.find((l) => l.header === 'Docker Version' && !l.recommendation);
+        return Promise.resolve(dockerIsFine);
```
Jaguar0625, line 162: can you explain when you wouldn't want to check docker?

src/service/VerifyService.ts
```diff
@@ -161,40 +175,76 @@ export class VerifyService {
...
+}
 
-    public async loadVersionFromCommand(command: string): Promise<string | undefined> {
-        return this.loadVersion((await BootstrapUtils.exec(this.logger, command)).stdout.trim());
+export class VerifyService {
+    private readonly expectedVersions: ExpectedVersions;
+
+    public actions: VerifyAction[] = [];
+
+    constructor(private readonly logger: Logger, expectedVersions: Partial<ExpectedVersions> = {}) {
```
Jaguar0625, line 188: why is this a Partial. when would you want to override default requirements?

test/service/VerifyService.test.ts
```diff
@@ -16,26 +16,34 @@
...
     });
+});
+
+describe('VerifyService', () => {
+    const currentNodeJsVersion = process.versions.node;
+    async function getCurrentVersions() {
+        const appVersionService = new AppVersionService(logger);
+        const currentDockerVersion = await appVersionService.loadVersionFromCommand('docker --version');
+        const currentDockerComposeVersion = await appVersionService.loadVersionFromCommand('docker-compose --version');
+        expect(semver.valid(currentNodeJsVersion, AppVersionService.semverOptions));
```
Jaguar0625, line 38: you seem to be testing  that `process.versions.node` is validly formatted. at a minimum, shouldn't this come from the src code?

src/service/VerifyService.ts
fboucquez: The exception is for misconfigured actions.  A well-configured action may have the issue that the command is not returning any version (so the possible undefinded)

https://github.com/fboucquez/symbol-bootstrap/blob/cbe72d35b05b6d692ba20041a3d08afb65d812c1/src/service/VerifyService.ts#L63

src/service/VerifyService.ts
fboucquez: I was thinking that souldRun could be an async test. There is no case so far so back to plain boolean

src/service/VerifyService.ts
fboucquez: For unit testing

src/service/VerifyService.ts
fboucquez: When docker is not installed (previous test), checking that docker can run as regular user is ignored

src/service/VerifyService.ts
fboucquez: Refactored common string

test/service/VerifyService.test.ts
fboucquez: Loading process.versions.node in test from src code

PR #399, opened 2023-04-25
[bootstrap] task: add option to configure node metadata

cryptoBeliever:
## What's the issue?

Newly added endpoint /node/metadata in REST returns meta-information about the node. Bootstrap doesn't allow users to define custom values for this endpoint.

## What's the solution?

Add the possibility to define node meta-information via custom preset. This can be done by:

```
gateways:
-  restNodeMetadata:
    key11: "value1"
    key21: "value2"
    key31: "value4"
    key51: 1
    key61: 10
```

### DIFF
--- a/presets/shared.yml
+++ b/presets/shared.yml
@@ -227,6 +227,8 @@ restProtocol: HTTP
 restSSLPath: '/symbol-workdir'
 restSSLKeyFileName: 'restSSL.key'
 restSSLCertificateFileName: 'restSSL.crt'
+restNodeMetadata:
+  _info: "Node metadata"
 statisticsServicePeerFilter: ''
 statisticsServicePeerLimit: 50
 statisticsServiceRestFilter: suggested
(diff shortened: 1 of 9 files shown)


presets/shared.yml
```diff
@@ -223,6 +223,8 @@ restProtocol: HTTP
 restSSLPath: '/symbol-workdir'
 restSSLKeyFileName: 'restSSL.key'
 restSSLCertificateFileName: 'restSSL.crt'
+restNodeMetadata:
+  _info: "Node metadata"
```
Wayonb, line 231: Do you know what's the default use case for this endpoint?
It would be nice to autofill the default values.

Wayonb (dismissed):
Looks good 👍🏾

presets/shared.yml
cryptoBeliever: For example, at this moment, additional information about the node listed in https://symbol-tools.com/symbolTools/view/tool/nodeList.html are added manually via an additional configuration file (https://symbol-tools.com/nodeSetting.json generated here https://symbol-tools.com/symbolTools/view/tool/nodeAdminManual.html). This is possible if you host the node, but Bootstrap will now support it. 
I don't know what default keys/values should be set. Default values may be wrongly read as set by the operator  :thinking:

Wayonb (approved):

PR #400, opened 2023-05-05
[bootstrap] fix: add stop_grace_period to server and broker containers

cryptoBeliever:
## What's the issue?

The default timeout for grace shutdown is 10s. In some scenarios, the server needs more time to shut down gracefully. That's why the server is shut down incorrectly, which could corrupt the data and will leave a .lock file.

## What's the solution?

Set stop_grace_period for:
- server to 300s
- broker to 20s


src/service/ComposeService.ts
```diff
@@ -200,6 +200,7 @@ export class ComposeService {
                         image: presetData.symbolServerImage,
                         command: serverCommand,
                         stop_signal: 'SIGINT',
+                        stop_grace_period: '300s',
```
Wayonb: Should this be configurable from the custom preset file?

src/service/ComposeService.ts
```diff
@@ -227,6 +228,7 @@ export class ComposeService {
                                     command: brokerCommand,
                                     ports: resolvePorts([{ internalPort: 7902, openPort: n.brokerOpenPort }]),
                                     stop_signal: 'SIGINT',
+                                    stop_grace_period: '20s',
```
Wayonb: Same here.  good be to be configurable.

Wayonb (changes requested):

src/service/ComposeService.ts
cryptoBeliever: Do you think someone would like to increase it even more? OK, I will add it to be configurable.

cryptoBeliever: Now it's possible to overwrite defaults. You can also define it for other services (not only broker and server) if it will be needed via custom preset.

src/service/ComposeService.ts
cryptoBeliever: Now it's possible to overwrite defaults. You can also define it for other services (not only broker and server) if it will be needed via custom preset.

presets/shared.yml
```diff
@@ -96,6 +96,10 @@ maxWriterThreads: 8
 maxDropBatchSize: 10
 writeTimeout: 10m
 
+# stop grace periods
+nodeStopGracePeriod: '300s'
```
Wayonb: dont need the single quotes

Wayonb (dismissed):
looks good 👍🏾

presets/shared.yml
cryptoBeliever: fixed :+1:

Wayonb (approved):

